PR #2, opened 2016-12-16
allow partial-dollar donations

x110dc:
#### What's this PR do?

Allows people to donate some fraction of a dollar like $10.50 and allows them to specify zero cents too (like 10.00 instead of just 10). 

Even though we only allow people to give whole dollar amounts at the support portal some people were changing the amount once they reached the checkout page to either specify zero cents or a partial dollar value. This was breaking. The PR fixes that case. 

- `make backing`
- `make interactive`
- `C_FORCE_ROOT=True celery -A app.celery worker --loglevel=INFO &`
- `python3 app.py`
- Visit http://local.texastribune.org/circleform?amount=84.01&installmentPeriod=monthly&installments=36 and complete the donation
- Confirm there are no errors and that you see 36 * 84.01 (3024.36) as the total amount donated in the console.
- Visit http://local.texastribune.org/memberform?amount=40.01&installmentPeriod=monthly and complete the transaction
- Confirm that there are no errors and that you see that '40.01' was donated on the console. 
- Confirm in both cases that you see an alert in #stripe with the correct amounts.

Yes.

#### Are there performance implications? If adding JS can it be done async? Do images/CSS/JS have cache headers set?

No.

Off-sprint bug found by Terry.

Let Terry and membership know that donating partial-dollar values is now supported, at least in the checkout app.

#### Next steps?

None.

#### Smells?

Not that I know of.

Not necessary.

### DIFF
--- a/salesforce.py
+++ b/salesforce.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+from decimal import Decimal
 import json
 import locale
 from pprint import pprint   # TODO: remove
@@ -21,6 +22,8 @@ from check_response import check_response
 zone = timezone(TIMEZONE)
 
 locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')
+
+TWOPLACES = Decimal(10) ** -2       # same as Decimal('0.01')
 
 WARNINGS = dict()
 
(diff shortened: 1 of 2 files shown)


AndrewGibson27:
LGTM. All the manual tests worked as expected.

PR #4, opened 2017-02-17
add note/styles to let users know they will be prompted for billing e…

brickaa:
Adds a note to let users know they'll be prompted for the billing address later. This should clear up confusion about users putting in the billing email address instead of the address for the person they are setting up the subscription for.

image

Per this task: https://3.basecamp.com/3098728/buckets/736178/todos/379371725


x110dc:
LGTM

PR #6, opened 2017-04-12
Make yearly/monthly donations input editable

AndrewGibson27:
#### What's this PR do?
Makes it so the box on `memberform` showing the dollar amount is editable like it is on `donateform`. This is a request from Terry, Rebecca and Erin.

To give the user more choice. It was decided this would be best for SMD 2017.

Go to `local.texastribune.org/memberform?installmentPeriod=monthly&amount=500` and verify the amount input box is editable. Try entering a non-numeric value and verify the form does not let you submit payment.

Try this with `installmentPeriod=yearly` as well -- and if you have time, with a partial-dollar amount.

No.

NA.

No.

Off-sprint SMD request.

They'll have more choice.

#### Next steps?
None.

#### Smells?
None.

#### TODOs:
None.

NA.

#### Technical debt note
Same.

### DIFF
--- a/templates/includes/display_levels_js.html
+++ b/templates/includes/display_levels_js.html
@@ -1,3 +1,7 @@
 <script>
   window.onload = display_level();
+
+  $('#amount').keyup(function() {
+    display_level();
+  });
 </script>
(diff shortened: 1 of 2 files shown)


x110dc:
There's some history here. The form used to be editable but then if someone changed amount to be above the circle threshold then suddenly they were on the wrong page (memberform vs. circleform).  If membership is ok with that now then cool. 

Also, the level doesn't change when I change the amount.

AndrewGibson27:
@x110dc Erin confirmed membership is OK with the circle threshold issue. And I made it so the level changes.

x110dc:
LGTM.

PR #7, opened 2017-05-16
Apple Pay

AndrewGibson27:
#### What's this PR do?
1. Integrates Apple Pay via Stripe per [these instructions](https://stripe.com/docs/apple-pay/web).
2. Refactors the checkout JavaScript.

1.The more ways people can donate, the better.
2. There were lots of globals hanging around, so I thought it would be good to clean up and moduralize the checkout JS.

Some caveats:
- This, of course, only works on Safari for Mac Sierra or higher.
- Despite following these [sandbox instructions](https://developer.apple.com/support/apple-pay-sandbox/) and these [Handoff instructions](https://support.apple.com/en-us/HT204681), I wasn't able to get my Mac to use the test card stored on our test iPhone. Instead, I had to add the test card to my Mac itself (System Preferences --> Wallet and Apple Pay).
- Apple Pay requires HTTPS, even locally, so you'll need something like [ngrok](https://ngrok.com/).

1. Go to `/donateform` and make sure you see the Apple Pay button.
2. Enter an invalid amount. Verify you get an error message when clicking either the "Pay by card" or Apple Pay button.
3. Leave either the first name or last name field empty. Verify you get an error message when clicking either the "Pay by card" or Apple Pay button.
4. Enter valid information and click the Apple Pay button. Confirm the prompt appears with the correct amount and card information. Make sure it asks for your email. Hit submit and confirm the charge goes through successfully.

Repeat steps 3-4 on `/memberform` and `circleform` with a variety of different query parameters. For example:
- `/memberform?amount=120&installmentPeriod=monthly`
- `/memberform?amount=999&installmentPeriod=yearly`
- `/circleform?&amount=209&installmentPeriod=monthly&installments=36`
- `/circleform?&amount=5000&installmentPeriod=yearly&installments=3`

Lastly, try regular old "Pay by card" checkout a few times, just to verify I didn't screw anything up.

No.

Well, we should test it on mobile. But that would require putting it on staging, which would in turn require that we put staging behind https. Is that possible?

No.

https://3.basecamp.com/3098728/buckets/736178/todos/469121094

It should be self-explanatory.

#### Next steps?
See to-dos.

#### Smells?
NA.

#### TODOs:
1. See if we can get a Mac to read a phone's cards via Handoff.
2. Test on staging.

I put in a Basecamp task to write docs for this. I want to wait until we see whether this is going to work via Handoff.

#### Technical debt note
A little more.


AndrewGibson27:
FWIW, the tests passed for me locally.

x110dc:
I confirmed most of the steps above, but of course, it tries to use my real credit card. Then it fails. I think that's the be expected at this point. 

It didn't ask for my email but it already had it listed.

x110dc:
If you want to dig into Flask work on Step 4 while I'm out go for it!  I got that part working locally but it's not working on test.

PR #10, opened 2017-08-09
Blast update

brickaa:
This PR removes Bourbon/Neat and updates the styles on all check out pages with our grid styles. Coming soon: some additional changes to The Blast.

More on the updates here: https://3.basecamp.com/3098728/buckets/2687322/todos/626163721#__subscriptions

For testing, review the following links
/donateform?amount=100
/memberform?installmentPeriod=yearly&amount=500
/memberform?installmentPeriod=monthly&amount=50
/circleform?installmentPeriod=yearly&amount=500
/blastform
/cfdo (i.e. the error page)

Also confirm the check-out view for both regular check-out and the blast by submitting stripe 
(4242 4242 4242 4242 + future data 123 security code)


AndrewGibson27:
These styles look grand! I think it's worth doing a cursory browser test on this, just to cover our bases. Also, might be good to open this on iPhone Safari and verify the Apple Pay button is working properly and looks OK. Other than that, I think it's ready to ship.

Lastly, this might be a separate PR thing, but on the Blast form, when the payment modal appears, I think it's supposed to say the dollar amount next to "Pay" in the button. Looks like the JS isn't getting the value properly in this line:
https://github.com/texastribune/salesforce-stripe/blob/master/templates/includes/blast_stripe_checkout_js.html#L32

AndrewGibson27:
I got Apple Pay running on my machine, and the buttons look good. Still might be worth putting this branch on staging, though, to verify it's working properly. Apple Pay doesn't work locally. I'll leave that up to you!

brickaa:
We tested on staging, all good

PR #11, opened 2017-08-15
Use nonprofit formula for Blast's fees value

AndrewGibson27:
Stripe gives us the nonprofit rate for Blast fees, as well. So people buying a subscription and checking the "agree to pay fees" box have been getting charged at that lower rate. However, the formula on the front-end has still been showing the fees amount at the original, non-nonprofit rate. This PR fixes that.

(Nothing serious -- it means people have actually been getting charged less than they expected.)

Testing:
- On `/blastform`, check that the fee amount for each option in the "subscription plan" dropdown is the same as when calculated with [this formula](https://github.com/texastribune/salesforce-stripe/blob/master/batch.py#L47-L65).
- Verify that if you select a different option from the dropdown, the fees value updates.

### DIFF
--- a/static/js/blast.js
+++ b/static/js/blast.js
@@ -31,20 +31,12 @@ var listen_for_installments = function() {
   });
 };
 
-var pay_fee_amount = function() {
-  var input_amount = $('select[name="amount"]').val();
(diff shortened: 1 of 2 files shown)


brickaa:
LGTM!

PR #12, opened 2017-08-22
Blast

brickaa:
What this PR does
+ Changes the Blast subscription form from a SelectField to RadioField
+ Updates the relevant scripts to account for the template change in the form
+ Fixes a javascript bug that didn't catch new value when the form subscription field changed

How to test:
+ Check that the styles look good on various sizes/screen dimensions
+ Check that when you switch values, the pay fees recalculates correctly and the Stripe processor picks up the correct value
+ Make sure you can still submit


templates/blast-form.html
```diff
@@ -80,8 +86,16 @@
       listen_for_fee_check();
     };
 
-    $('select[name="amount"]').change(function() {
+    $('input[name="amount"]').click(function() {
```
AndrewGibson27: Do we need this logic? If they're radio buttons, clicking one should automatically deselect the others. You should be able to use the `.change` listener for this.

In any case, if you keep this block, I'd recommend updating it to use some jQuery-specific stuff to ensure cross-browser compatibility:
- `$(this)` instead of `this`
- `$(this).attr(ATTR_NAME, ATTR_VALUE)` instead of `setAttribute/removeAttribute`
- `$(this).prop('checked, true/false)` instead of `this.checked`

AndrewGibson27:
Left one comment. But everything WFM. Maybe @x110dc should sign off on the change to the Python?

AndrewGibson27:
Addendum: I did some cursory browser testing in IE and Edge, and it looks like the $325 option is getting some weird font sizes.

AndrewGibson27:
screen shot 2017-08-22 at 5 58 20 pm

brickaa:
thanks @AndrewGibson27 -- addressed those bits of feedback.

@x110dc can you do a sanity check that this doesn't break stripe?

x110dc:
Sorry, I took so long to get to this. The Python bits look fine to me.

PR #14, opened 2017-08-30
disable CSRF

x110dc:
#### What's this PR do?

Disables CSRF.

Every once in a while a user will get a CSRF error. This avoids that. It's not doing us any good anyway since there's no session in this app.

Try any URL for the app and make sure you don't get a CSRF error in the log. For example:

http://local.texastribune.org/memberform?campaignId=blah&amount=85

No.

No.

No.

Off sprint task.

N/A

#### Next steps?

N/A

#### Smells?

None.

No.

#### Technical debt note

Same.

### DIFF
--- a/config.py
+++ b/config.py
@@ -8,12 +8,14 @@ def bool_env(val):
     """Replaces string based environment values with Python booleans"""
     return True if os.environ.get(val, False) == 'True' else False
 
+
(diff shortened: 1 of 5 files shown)


AndrewGibson27:
WFM.

PR #16, opened 2017-10-04
show one-time option on monthly form, fix styles

brickaa:
This allows the user to switch to a "one-time" donation from the monthly donation URL — just as they can for the yearly donation. While we don't want to encourage one-time donations, it's better to have a user give one-time than drop off entirely because they don't want to set up a recurring charge and can't figure out how to get to a one-time donation place, which they can't find at all from this URL right now.

I also fixed the dollar sign and input field, so that it expands to fill the available width and doesn't go to two lines when the screen shrinks.

To test: 
Go to a monthly donation url `/memberform?installmentPeriod=monthly&amount=15` and make sure you can process a one-time donation

Resize and make sure that the amount input looks good on all sizes

### DIFF
--- a/forms.py
+++ b/forms.py
@@ -25,7 +25,7 @@ class BaseForm(FlaskForm):
 
 class MemberForm(BaseForm):
     openended_status = RadioField(u'Membership Duration',
-        choices=[('Open', 'Sustaining'), ('None', 'One Year')],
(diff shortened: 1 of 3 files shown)


AndrewGibson27:
Everything WFM. I'd like to check w/ @x110dc to ensure this won't break anything from a Salesforce standpoint.

x110dc:
LGTM.

PR #17, opened 2017-11-21
Ensure 1password doesn't change amount on autofill

AndrewGibson27:
#### What's this PR do?
Prevents 1password's credit-card autofill feature from changing the amount a user originally entered. There's weirdness between 1Password and the Stripe modal: Whatever card-expiration month the former fills in also gets dropped into the Flask form's `amount` input.

Anna H. reported this and sent along some code a VOSD developer wrote to fix it. It seemed overly complicated, though: I'm simply setting the amount field to `readonly`when the Stripe modal is open. It seems to work.

To ensure people don't give the wrong amount.

- `make backing`
- `make interactive`
- `python3 app.py`
- Add a credit card to 1password and ensure you have the browser extension installed
- On both `/donateform` and `/memberform`, fill out the form as normal, then click "Pay By Card." Use 1password to autofill the credit-card information. Confirm that doing so does not change the input amount in the form below the overlay. Close the Stripe modal and confirm the amount field is editable again.
- On `/circleform`, once again fill out the form, hit "Pay by Card," and autofill the credit card info with 1password. Confirm the amount in the Flask form does not change. Close the Stripe modal and confirm the amount field is still read-only.
- Try checking out on `/blastform` and `/blast-vip`, just to ensure nothing seems off. (I tested there, too.)
- If an iPhone user with the 1password app installed can try this (this branch is on staging), that'd be grand. I'll take care of Android.

No.

I tested successfully on desktop Chrome, Firefox and Safari. I feel pretty good about those results.

No.

https://3.basecamp.com/3098728/buckets/736178/todos/784219601

NA.

#### Next steps?
NA.

#### Smells?
Toggling read-only on the amount input is a slight smell, but it's not much code, either.

#### TODOs:
NA.

NA.

#### Technical debt note
A tiny bit more.

### DIFF
--- a/templates/blast-form.html
+++ b/templates/blast-form.html
@@ -21,7 +21,7 @@
       </div>
     </div>
   </header>
-  <div class="grid_container blast_main">
+  <div id="blast" class="grid_container blast_main">
     <form action="/submit-blast" method="post" class="blastform grid_separator--l">
       <input name="campaign_id" id="campaign_id" value="{{ campaign_id  }}" type="hidden">
       <div class="form-outer-section outer-section narrow">
--- a/templates/blast-vip.html
+++ b/templates/blast-vip.html
@@ -21,7 +21,7 @@
       </div>
     </div>
   </header>
-  <div class="grid_container blast_main">
+  <div id="blast-vip" class="grid_container blast_main">
     <form action="/submit-blast-vip" method="post" class="blastform grid_separator--l">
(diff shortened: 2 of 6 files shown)


x110dc:
I did all the testing you describe except the iPhone part. I don't think there's an automatic way to fill in credit card details with the 1Password app on iPhone. 

LGTM

PR #18, opened 2017-11-28
fix for multiple emails being sent about the same duplicate account

x110dc:
#### What's this PR do?

Adds a fix for duplicate emails being sent out for the same account.

To stop annoying Anna and the others who read and process those emails.

There's not really a way to manually test it, but I wrote a unit test so run `make test` and make sure it works.

Yes.

N/A

No

https://3.basecamp.com/3098728/buckets/736178/todos/784339426

Tell Anna that it's fixed.

#### Next steps?

None.

#### Smells?

No.

No.

#### Technical debt note

Less

### DIFF
--- a/salesforce.py
+++ b/salesforce.py
@@ -61,7 +61,7 @@ def send_multiple_account_warning():
     Send the warnings about multiple accounts.
     """
 
-    for email in WARNINGS:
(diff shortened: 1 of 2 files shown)


AndrewGibson27:
WFM.

PR #20, opened 2017-12-20
set temporary redirect on blast-vip to standard checkout

brickaa:
This PR comments out the route for the `blast-vip` promo page and sets a 302 redirect to the main `blastform` checkout.

This promotion ends on Dec. 31, so we no longer want people who come to this page to be able to sign-up for The Blast at the discounted rate. 

I left the code in the file, but commented out, as I expect we'll resurrect this promotion or another one in the future.

To test —
Go to local.../blast-vip and make sure that it redirects to https://checkout.texastribune.org/blastform with a 302.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1,7 +1,7 @@
 import os
 import sys
 
-from flask import Flask, render_template, request, send_from_directory
+from flask import Flask, redirect, render_template, request, send_from_directory
 from forms import DonateForm, BlastForm, BlastVIPForm, MemberForm
 from raven.contrib.flask import Sentry
 from sassutils.wsgi import SassMiddleware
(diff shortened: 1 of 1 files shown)


AndrewGibson27:
WFM. Left one comment.

app.py
```diff
@@ -150,19 +150,22 @@ def submit_blast():
...
-        amount = 275
-    installment_period = request.args.get('installmentPeriod')
-
-    campaign_id = request.args.get('campaignId', default='')
-
-    return render_template('blast-vip.html', form=form,
-            campaign_id=campaign_id, installment_period=installment_period,
-        openended_status='Open', amount=amount,
-        key=app.config['STRIPE_KEYS']['publishable_key'])
+    return redirect("https://checkout.texastribune.org/blastform", code=302)
```
AndrewGibson27: Might be worth making this a relative URL so that it redirects you to the local `blastform` when you're in dev mode.

PR #22, opened 2018-01-24
Create "/donate" for homepage test

brickaa:
While more than 60% who initiate checkout complete it, in total only 5-7% of people who click on "Donate" or "Support Us" from the homepage currently complete a membership sign-up. That means we're losing a very significant number of people on the support.texastribune.org landing page. We'd like to test whether sending them directly to checkout will significantly improve membership conversion rates.

This PR builds a new URL — `/donate` — with a new form similar to `MemberForm()` called `MemberForm2()`. Rather than accept the installment_period (yearly/monthly/one-time) from the URL parameters, it allows the user to select that on the form itself. 

To test, go to `/donate?amount=60` and submit various types of memberships (change amount, membership type, etc.) and attempt to break the form by leaving out each of the required fields. For ref, here's the Stripe [testing card list](https://stripe.com/docs/testing#cards).

TODOs:
- [x] Bug — putting the installmentPeriod parameter in the URL should load the page with that membership type checked. It's not currently doing that.
- [ ] Need to test that the form correctly submits the data to salesforce. Manual tests work, but I'm not sure how it's coming out the other end yet.
- [x] Review Design w/ art
- [ ] Review copy/design with Quarls


AndrewGibson27:
This is a swell-looking page! Overall, I'm 100 percent in favor of moving toward consolidating the support and checkout apps. I'll do a code review when the WIP is gone, but some initial (more philosophical) thoughts:
+ I'm curious what the best way is for this to coexist with the support app. It seems like having the full form widget in the checkout app almost renders the support app irrelevant (not a bad thing!). But, during FMD other campaigns, the latter is where we've been building campaign-specific pages -- and thus where Rebecca sends people in her marketing materials. Would the checkout app start getting campaign pages instead? Or is `/donate` purely for linking to from the home page?
+ Should we get Emily Y. involved in this code? Might be a good intro to the checkout app, plus this is (sort of) audience stuff.
+ If you want to test your hypothesis without building this full-on page, you could maybe just drop a few links on the home page that go to existing pages in the checkout app and use query parameters. For example, maybe a row of three links: "Donate $15 once", "Donate $20 a month", "Donate $100 a year". Maybe a stupid idea -- just spitballin' here.
+ As with anything in this app, I defer to @x110dc on whether this will impact Salesforce data collection at all.

brickaa:
As I was building this, I went a little bit further down the merge with donations-app path than originally planned.  I wanted to make the test as much like support.tt.org as possible, so we could really isolate how clicking through to another page impacted drop-off rates. While we could just send people to the current checkout pages, the design is so different that I worry it could impact the test results. My plan is to create this page for the test and then if we decide to kill it, set up a temporary redirect back to support.tt.org until we do an officially merge/move of donations-app and salesforce-stripe. If we decide not to kill it, because the test shows a positive impact on conversions, then we'll schedule a larger discussion on what else needs to be brought over (navigation bar? other pages/links? campaigns? etc.) and the timeline for doing all of that. We also had a conversation today about adding group sign-up functionality for The Blast, which would also be a big impact on the salesforce-stripe app, and will need to be part of that conversation.

Bringing in Emily Y is a great idea. When I sked a meeting w/ quarls and ben i'll invite her too.

AndrewGibson27:
Some things to note:
+ The current `/memberform` displays a donor's level as the amount is adjusted. Rebecca & co. might want that here, too.
+ It's worth giving `/circleform` a whirl to confirm everything there is working well.
+ After this is deployed, we should definitely donate $1 via Apple Pay just to confirm it's working (seems to be locally, but it's virtually impossible to guarantee that in a non-production environment)
+ Also worth double checking the fee amount is still being calculated correctly.

Looking good, though!

static/js/all.js
```diff
@@ -82,6 +82,13 @@ var listen_for_installments = function() {
   });
 };
 
+var listen_for_installment_period = function() {
```
AndrewGibson27: Can be deleted, right?

templates/includes/change_installment_period_js.html
```diff
@@ -0,0 +1,20 @@
+<script>
+  $('input[name="installment_period"]').click(function() {
+    var installment_period_id = this.id;
+    var openended_status_open = document.getElementById("openended_status-0");
+    var openended_status_none = document.getElementById("openended_status-1");
+
+    console.log(installment_period_id);
```
AndrewGibson27: Can kill.

templates/includes/change_installment_period_js.html
```diff
@@ -0,0 +1,20 @@
+<script>
+  $('input[name="installment_period"]').click(function() {
```
AndrewGibson27: If you totally wanted to ditch jQuery, you could use `document.querySelectorAll` (certainly not necessary). Also, I think we should technically be binding to the `change` event (not click), but it shouldn't really make a difference.

templates/includes/change_installment_period_js.html
AndrewGibson27: Delete

AndrewGibson27:
@brickaa Scroll up a bit for a bulleted list of questions/notes.

brickaa:
@AndrewGibson27 @x110dc FYI, this is ready for review!

AndrewGibson27:
LGTM. I pushed a commit to remove a print statement in `app.py`. Not sure how much/if this needs browser testing -- I'll leave that up to you. Would still definitely love a final sign-off from @x110dc to make sure the form data is being sent to the Salesforce logic properly. I've checked that multiple times, but hey, I'm paranoid.

x110dc:
LGTM.

PR #23, opened 2018-01-26
Small JS fixes

AndrewGibson27:
#### What's this PR do?
+ Fixes bug where an empty value in the amount field caused "NaN" to appear for the fee amount
+ Hides the "level" part of the form if an installment period is not specified as a query parameter

Bugs are bad.

+ Go to `/donateform` and clear out the amount field. Confirm the fee amount goes blank.
+ Go to `/memberform?amount=50` and confirm there is no part of the form that says "level."
+ Go to `/memberform?amount=50&installmentPeriod=monthly` and confirm there _is_ a "level" part of the form. Confirm its text updates as you change the amount.

No.

NA.

No.

https://3.basecamp.com/3098728/buckets/736178/todos/848429970

NA.

#### Next steps?
NA.

#### Smells?
TBH all the JS in this app is a smell.

#### TODOs:
NA.

NA.

#### Technical debt note
Same.

### DIFF
--- a/static/js/all.js
+++ b/static/js/all.js
@@ -22,7 +22,7 @@ var display_level = function() {
     } else {
       level_label.text('');
     }
-  } else {
+  } else if (frequency == 'yearly') {
     // detemine level and update text based on yearly frequency
     if (input_amount == 10) {
       level_label.text('Student');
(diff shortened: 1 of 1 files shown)


x110dc:
LGTM.

PR #26, opened 2018-04-09
SASS cleanup + text updates 

emilyyount:
**What's this PR do?**
- Updated, streamlined text from Amanda. 
- Updated SASS /utils to match texastribune repo. Exception is the _typography.scss file because the fonts are loaded differently in this repo. 
- Added /components/_forms.scss from texastribune repo.
- Updated _new_theme.scss to rely more on /utils and /components. Also general cleanup that goes along with the streamlined text. I passed classes to the flask forms so we don't have to write CSS on the base tags. 
- Created separate include HTML files for SMD. Those are commented out in this PR, enabling us to deploy this update if we want. 
- Removed the background image, making the text more readable and the focus on the checkout form. Waiting on art for a background image that is either specific to SMD or can be used on this page going forward. 

**Why are we doing this? How does it help us?**
- Reduce the amount of reading you have to do to understand what it means to be a member.
- Clean up SASS on the road to consolidating support and checkout apps. 

**How should this be manually tested?**
- I need to do more testing if we are going to deploy this before SMD. I haven't looked at anything other than the landing page since we were planning to redirect those during the drive. Maybe those aren't affected since I stuck to the _new files? 

**Have automated tests been added?**
No

**Has this been tested on mobile?**
Only in the emulator. I need to test on my device. 

**Are there performance implications?**
I left all Javascript alone. 

**What are the relevant tickets?**
I need to add line items, but this is the Basecamp project: https://3.basecamp.com/3098728/buckets/6964063/todosets/965718562

**How should this change be communicated to end users?**
We need to show this to Amanda. Also not sure the copy has been approved by Emily and Evan.

**Next steps?**
Find out if we want to wait on art to make a new background illustration. Run this by Amanda, et al. 

**Smells?**
The fonts are loaded differently than the texastribune repo. I'm not sure if that's an old holdover or a new way of doing it. They are loaded in new_layout.html instead of in the _typography.scss file.


static/sass/_smd.scss
```diff
@@ -0,0 +1,37 @@
+.grid_separator--xl {
```
brickaa: can this be added to the `._grid.scss` file?

static/sass/_smd.scss
```diff
@@ -0,0 +1,37 @@
...
+
+  @include mq($until: bp-s) {
+    &:first-child {
+      margin-bottom: $size-l;
+    }
+  }
+
+  &--amount {
+    @extend .smallcaps;
+    font-size: $size-s;
```
brickaa: for consistency, we tend to alphabetize styles. you can quickly resort them by highlighting the group and hitting F5

static/sass/components/_forms.scss
```diff
@@ -0,0 +1,216 @@
+%input_text {
+  @extend .font-sansserif;
+  border: 1px solid $color-gray-light;
+  height: 27px;
```
brickaa: I try to avoid magic pixel numbers (but it's not always possible and this may be an exception). Could the heights on this and `%input_text--l` be set as variables? Maybe we need an update to the `utils/_forms.scss` for input heights, and could standardize it in the main repo too with a variable.

static/sass/components/_forms.scss
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  fieldset {
+    border: 0;
+  }
+
+  &_labels {
+    padding-top: $size-b;
+
+    li {
```
brickaa: This is my fault (because I originally wrote this) but it's kinda smelly to put so many styles on the base elements inside a form, even with a class... I did this because there were so many form templates to rewrite & didn't want to bother adding classes to all the templates. Not super important, but we could work on refactoring that here since its in a smaller context and later pull those updated styles/templates into the main repo.

static/sass/utils/_buttons.scss
```diff
@@ -60,6 +71,7 @@
 // dark buttons in mobile nav are a bit shorter so nav doesn't change height on scroll
 $button-nav-height: 24px;
 $button-default-height: 26px;
+$button-large-height: 44px;
```
brickaa: do these match up with the form input heights? perhaps we could standardize them

AndrewGibson27:
This is all spectacular and delightful. Huge fan. A few thoughts:
+ As far as Sass organization goes, I'm a fan of having a `pages/` subdirectory, and within that, a file that imports all the stuff necessary for that particular page. So here, we could have `pages/donate.scss`, `pages/blast.scss` and, eventually, `pages/circle.scss`. Is that overkill for this small of an app? IDK. Maybe the one stylesheet to rule them all approach is best.
+ We should think about using CSS grid for this -- maybe not now, but soon. [Browser support](https://caniuse.com/#feat=css-grid) is pretty good these days, and if we could use a vendor library based on it, it would save us the trouble of having to maintain our own system and write flexbox hacks around stupid IE 11.
+ When I rewrite the JS, the form markup is all going to be generated client side. That of course doesn't change anything about the styles -- we'll just need to apply the classes inside a Vue.js component file. I can help with that.

static/sass/components/_forms.scss
emilyyount: @brickaa Totally agree. I think we'll be able to address this after Andrew's refactor because all of the HTML won't be buried in the wtf forms. Okay if we hold off on this fix until after that?

cc @AndrewGibson27

static/sass/utils/_buttons.scss
emilyyount: Yes, this matches with the form heights. I agree about standardizing. I was nervous to update everything since it affects so many pages. I think we should refactor after we have killed off some of the extra routes.

emilyyount:
@brickaa Thank you for the review! I added a thumbs up to items that I have fixed in the repo and am about to push. I wasn't sure what you meant about the alphabetize part, so we can talk about that on Monday. 

----

@AndrewGibson27 That all sounds great to me. 

For now, I only made changes to `_new_theme.scss` because that file only affects checkout.texastribune.org/donate. I was nervous to make more structural changes to the SASS because we aren't in a position to update the other routes before SMD. Once we get around to killing off the extra routes, I think we could do a more thorough refactor of the SASS, based around the routes we plan to support. 

That is just one way of thinking about it though, and I'm not well-versed in merging repos together, etc. I'm totally open to a different approach. 

I'd also like to make a few updates to _forms.scss in the texastribune repo, and have those carry over to these pages ... I get a little overwhelmed thinking about where to start with that though. Would love guidance on how to pace out these updates.

AndrewGibson27:
@emilyyount Totally, totally understand. No need to refactor anymore before SMD -- don't want your head to explode!

As far as the alphabetization thing, I think @brickaa is talking about having the properties in your selectors sorted that way -- i.e.:
```css
/* bad */
.foo {
  position: relative;
  font-size: 18px;
}

/* good */
.foo {
  font-size: 18px;
  position: relative;
}
```
In Atom, I use the [sort-lines](https://github.com/atom/sort-lines) plugin. Not sure if there's a more efficient way to do it.

emilyyount:
@x110dc Thank you for the zip code updates! Amanda would like the zip code to be optional, now that we've moved it back to the form on the page. Sorry for the extra work!

x110dc:
@emilyyount I've added the backend parts for the ZIP Code. And this branch is up on https://stripe-testing.texastribune.org/ right now.

AndrewGibson27:
OK! Looking really good. Two things to note, then I think it's ready to ship.
1. Note that the JS in `stripe_checkout_js.html` is also used on the existing pages -- `/donateform`, `/memberform` and `/circleform`. That means the changes you made re: showing/hiding Apple Pay ([this stuff](https://github.com/texastribune/salesforce-stripe/blob/smd/templates/includes/stripe_checkout_js.html#L158-L170)) and preventing form submission ([this stuff](https://github.com/texastribune/salesforce-stripe/blob/smd/templates/includes/stripe_checkout_js.html#L57-L58)) also need to work there. Right now, it looks like it's not -- but that should be a simple fix.
2. If you don't have control over this, no big deal, but I noticed the Apple Pay logo is kinda small:
screen shot 2018-04-23 at 5 30 50 pm

emilyyount:
@AndrewGibson27 Thanks!

I pushed fixes for those issues. I know the size of the apply pay logo is fixed. I would appreciate a second set of eyes on the button change to make sure I didn't mess that up.

emilyyount:
@AndrewGibson27 Background-image should be better now. Tiled a 24KB image instead.

PR #28, opened 2018-04-17
SMD redirects + smart defaults

AndrewGibson27:
#### What's this PR do?
+ Temporarily redirects `/donateform` and `/memberform` to `/donate`, which will be the SMD campaign page.
+ Makes it so `/donate` doesn't 404 when an amount query parameter is not provided

Amanda request so they can better track money earned during SMD.

+ Visit both `/donateform` and `/memberform`. No matter the query params appended to each, they should redirect to `/donate?campaignId=70146000000M0pM`.
+ Visit `/donate` with no params. It should default to $10/month.
+ Visit with an `amount` param but not an `installmentPeriod` one. It should show the amount and default to monthly.
+ Visit with an `installmentPeriod=/monthly/yearly/once` param but not an amount param. If monthly, the amount should be 10. If yearly, 75. If once, 60.
+ Visit with a bogus `installmentPeriod` param and no amount param. It should default to $10/month.
+ Visit with a bogus `installmentPeriod` param but a valid amount param. It should show the amount and default to monthly.

No.

NA.

No.

Off-sprint SMD prep.

NA.

#### Next steps?
NA.

#### Smells?
The redirects are temporary, so meh.

#### TODOs:
* [x] Confirm amounts with Amanda.

NA.

#### Technical debt note
Same.

### DIFF
--- a/app.py
+++ b/app.py
@@ -89,21 +89,21 @@ def member2_form():
     if installment_period == 'monthly':
         openended_status = 'Open'
         if amount is None:
-            amount = 10
+            amount = '10'
     elif installment_period == 'yearly':
         openended_status = 'Open'
         if amount is None:
-            amount = 50
+            amount = '50'
     elif installment_period == 'once':
         installment_period = 'None'
         openended_status = 'None'
         if amount is None:
-            amount = 20
+            amount = '20'
     else:
         installment_period = 'monthly'
         openended_status = 'Open'
         if amount is None:
-            amount = 10
+            amount = '10'
     
     form.installment_period.default = installment_period
     form.openended_status.default = openended_status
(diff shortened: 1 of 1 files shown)


app.py
```diff
@@ -68,35 +72,39 @@ def member_form():
...
-        message = "The page you requested can't be found."
-        return render_template('error.html', message=message)
-
-    campaign_id = request.args.get('campaignId', default='')
     installments = 'None'
+    campaign_id = request.args.get('campaignId', default='')
     installment_period = request.args.get('installmentPeriod')
+    amount = request.args.get('amount')
 
     if installment_period is None:
```
tjb346: This could be simplified by removing lines 85 and 86 and changing line 82 to ```installment_period = request.args.get('installmentPeriod', 'monthly')```

tjb346:
WFM, see comment.

PR #33, opened 2018-05-17
Use Webpack to build stylesheets

AndrewGibson27:
#### What's this PR do?
Adds the proper configuration for Webpack to compile and build stylesheets.

My original intent was simply to add vendor-prefixing to Flask's CSS-build process. Along the way, I noticed static assets in this app have no cache-busting hashes attached to them. The `Cache-Control` header for assets is only one day, but that still means browsers might not pick up bug fixes for 24 hours. Given Webpack can handle both vendor prefixing and cache-busting hashes, it feels right to just start using it to build CSS.

Note that no production pages are using this build system yet. They'll start to as I roll out the revamped Vue.js-ified forms. But that PR is already big (and unrelated), so it felt better to get these updates in master. I'll deal with integrating them in production later.

1. Make sure you're not already inside the Docker container.
2. `make interactive`
3. `yarn run dev`
4. Go to `/devdonate` and view the page source. Confirm you see one stylesheet and three scripts. Each should have a hash before the file extension. Open in each in a new tab and confirm none 404.
5. Go to `/donate` and `/circleform` and confirm each looks normal.

No.

NA.

No.

No ticket, but related to the larger refactor of the checkout front-end code.

NA.

#### Next steps?
Integrate into Vue.js branch and production.

#### Smells?
The structure of `static/` will look much better when all the forms have been refactored. It's in a transitional state right now.

#### TODOs:
NA.

Again, the app is in a transitional state. I'll do a full documentation update when the refactor is done.

#### Technical debt note
Same.


x110dc:
LGTM.

PR #35, opened 2018-05-21
Refactor forms JS

AndrewGibson27:
#### What's this PR do?
+ Creates a system of Vue.js components for form-handling across the app
+ Integrates the new Stripe front-end code into this system (including support for the Payment Request API, Android Pay, Microsoft Pay and Apple Pay)
+ Uses these components to rebuild the form on `/donate`

+ A granular, component-based system will allow us to more easily create forms with different combinations of elements, as well as A/B test them
+ Using Vuex, Vue's store manager, we can store form data at a level "above" the form itself. This means information entered into one form (email address, for example) can be auto-filled on other forms elsewhere on a page.
+ The forms already require JS to work, so it feels cleaner to go all-in on a JS approach as opposed to have bad jQuery sprinkled throughout HTML
+ Vue.js is just great in general -- it's secure, has a large community and will make writing unit tests down the line much easier than random jQuery
+ The new Stripe front-end code is great (ditches the modal, adds support for all the native payments mentioned above)

+ Rebuild the container from scratch: `make interactive`
+ `yarn run dev`
+ Go to `/donate`. Try filling out the form and submitting with a variety of different input combinations. Confirm validation is working for email address, a valid 5-digit zip code, a numeric amount and valid card info.

Now, test that the query parameters are working as expected:
+ Visit /donate with no params. It should default to $10/month.
+ Visit with an amount param but not an installmentPeriod one. It should show the amount and default to monthly.
+ Visit with an installmentPeriod=/monthly/yearly/once param but not an amount param. If monthly, the amount should be 10. If yearly, 75. If once, 60.
+ Visit with a bogus installmentPeriod param and no amount param. It should default to $10/month.
Visit with a bogus installmentPeriod param but a valid amount param. It should show the amount and default to monthly.

No. One day ...

I've browser tested all the JS, but we'll want to test again when the styles are finalized.

More JS than before, but it's non-render-blocking and will benefit from Webpack's long-term caching capabilities. Overall, not an issue.

https://3.basecamp.com/3098728/buckets/736178/todos/1038482429

Hopefully they'll just adore it.

#### Next steps?
Bring to Circle and Blast pages.

#### Smells?
The `static/` directory is in a transitional state right now, just because some of the pages are still using the OG JS setup. When all pages have been converted, I'll probably put in a final PR to do some cleanup and re-organization.

#### TODOs:
* [x] Finalize layout
* [x] Browser test CSS
* [x] Test native payments

I will write comprehensive docs for this refactor once it's complete. I'm afraid writing something during this transitional state would just lead to confusion.

#### Technical debt note
Less.


x110dc:
LGTM

PR #36, opened 2018-05-22
Update contact phone number for group rates

mailbackwards:
New number for group rates. Make sure it shows up right. This is barely enough for a PR, but I haven't deployed this app before so I might need help with that part in particular.


x110dc:
LGTM. The deployment instructions are in the `README.md`. If you run into anything let us know. We can also deploy it for you if you need that.

PR #37, opened 2018-06-05
Create customer JSON endpoint for better client-side card validation

AndrewGibson27:
#### What's this PR do?
+ Creates the `/create-customer` JSON endpoint which takes an email address and Stripe token. If customer creation is successful, it returns the customer ID. If not, it returns a `CardError`, which is then shown to the user.
+ Refactors a bunch of JS logic to accommodate this change.

We got Paper Trail errors indicating users were getting payments rejected because of invalid CVCs. Unfortunately, Stripe 3 does not validate this client side. So we're now checking for card validity via the aforementioned JSON endpoint and using its response to show errors.

+ Rebuild from scratch: `make interactive`
+ `yarn run dev`
+ Go to `/donate`
+ Fill out the form besides the card input. Try to submit and confirm you get a red error below the card input saying "incomplete card input."
+ Now, enter a non-test card. After you complete entry, the submit button should temporarily become translucent to indicate token-fetching is in progress. After it finishes, the red error should change, this time noting that you're trying to use a live card in test mode.
+ Re-fill out the card input with the test card (`4242 4242 4242 4242`). Confirm the error disappears after token fetching has completed and that you can finish the form submission.

**EMAIL INSTRUCTIONS**
+ Make it so [this line](https://github.com/texastribune/salesforce-stripe/blob/ajax-customer/app.py#L174) checks some random string instead of the sent email address (or otherwise make the validation fail).
+ `yarn run dev`
+ Go to `/donate`
+ Fill out the form. You can use either a Stripe test card or a real one. After you finish the CVC entry, the email input above should get an error message about the server not accepting the email address. The card input should also have its contents cleared.
+ Fill out the email input with a new well-formatted address and confirm the error goes away.

No.

NA.

There's a small delay while we validate card input on the server, but I give users visual feedback (the translucent button) to indicate this is happening.

Off-sprint.

NA.

#### Next steps?
NA.

#### Smells?
I can't test that a good CVC-specific message will be returned until we deploy this to production. That's because, in test mode, the error message will always be about using a live card (not about invalid CVCs).

#### TODOs:
NA.

I added some inline comments.

#### Technical debt note
Probably a little more, but unavoidable.


app.py
```diff
@@ -178,29 +192,10 @@ def charge():
     customer_first = request.form['first_name']
     customer_last = request.form['last_name']
 
-    email_is_valid = validate_email(customer_email)
-
```
x110dc: Could it be an issue if we're no longer checking for a valid email server-side? Especially since this is now an API endpoint I'd hesitate to rely on the client checking. But maybe I'm missing something.

x110dc:
You can overcome your smell by using card number `4000 0000 0000 0101`. See here: https://stripe.com/docs/testing#cards-responses

app.py
AndrewGibson27: @x110dc I brought this back and updated the manual-testing steps above.

x110dc:
LGTM

PR #38, opened 2018-06-25
Minor updates to /donate 

emilyyount:
Per Amanda's email, I made a few updates to the checkout page. 

- Moved checkout form beneath membership sell.
- Moved festival discount info to membership benefits.
- Reordered form radio buttons (one time --> monthly --> yearly). 
- Added a button to the header that links to the checkout form 

Once this is published, I'll update the story page membership link to take readers straight to the checkout form. 

LMK if you see any issues!


templates/member-form2.html
```diff
@@ -7,124 +7,116 @@
...
+          </div>
+          <div class="smallcaps grid_separator">Members receive ...</div>
+          <ul class="list--bulleted">
+            <li>Behind-the-scenes access to the Texas Tribune newsroom, including Q&As with our reporters and sneak peeks at special projects.</li>
+            <li>$50 off <a href="https://festival.texastribune.org/">Texas Tribune Festival</a> badges, as well as priority entry to all panel dicsussions. Texas Tribune Festival is our annual, three-day meeting of the brightest minds in politics and policy — an opportunity to connect with people, like you, who care about the future of Texas. <a href="https://festival.texastribune.org/registration">Register here</a>.</li>
+            <li>Our utmost gratitude and a special place on our <a href="https://www.texastribune.org/support-us/donors-and-members/#2018">growing donor wall</a>.</li>
+          </ul>
+        </div>
+        <div id="join-today" class="donation_form section-padding">
```
AndrewGibson27: Is this ID used anywhere?

AndrewGibson27:
LGTM. I tacked on a commit to remove some compiled files that ended up in version control. It's my opinion that the yellow anchor button to the form should be larger, but I'll always defer to y'all and art on choices like that.

This is incredibly nitpicky, but I'd argue that for semantics, "Frequently Asked Questions" should be an `<h2>`, and that each section beginning with an `<h2>` should be its own `<section>` element inside of `<main>`.

But, pending art sign-off, I'd say deploy any time.

emilyyount:
@AndrewGibson27: Thank you for reviewing! 

I moved the FAQ into <main> and made it a <section>. I didn't move the form into it's own section because I prefer the member quotes to stack below the form on mobile. I also moved text into includes so it's easier to digest ... quotes.html, benefits.html and faq.html.

I will check on the size of the button. My hunch is that we will keep it small because we want to emphasize making the case over getting you to the form. In cases where we have already made the case, we will take you right to the form.

emilyyount:
@AndrewGibson27: I got approval from Art and Inks tweaked the tagline copy. 

Thanks for your help!!

PR #41, opened 2018-07-05
Do not call Stripe API if customer ID does not exist

AndrewGibson27:
#### What's this PR do?
+ Checks whether the `customerId` field exists in a form submission for retrieving a customer via Stripe.

To avoid throttling Sentry and Stripe.

+ `yarn run dev`
+ Fill out the form on `/donate` and confirm you can successfully submit it.
+ Change `customerId` on [this line](https://github.com/texastribune/salesforce-stripe/blob/customer-id/forms.py#L31) to something else.
+ Wait for Flask to restart, then submit the form again. Confirm you get an error/we're sorry page. (This is confirming that Flask does make it to [this statement](https://github.com/texastribune/salesforce-stripe/blob/customer-id/app.py#L259) if the form is invalid.)

No.

NA.

No.

Off-sprint.

NA.

#### Next steps?
I'll bring this same validation to the Circle form in the refactor PR.

#### Smells?
In general, the server-side form validation could be stronger in this app. I've left it alone during my refactor efforts and have instead focused on client-side validation. I'd like to handle the server-side stuff in a separate PR.

#### TODOs:
NA.

NA.

#### Technical debt note
Same.

### DIFF
--- a/app.py
+++ b/app.py
@@ -113,7 +113,7 @@ def circle_form():
 
 @app.route('/circle-charge', methods=['POST'])
 def circle_charge():
-    form = DonateForm(request.form)
+    form = CircleForm(request.form)
     pprint('Request: {}'.format(request))
 
     customer_email = request.form['stripeEmail']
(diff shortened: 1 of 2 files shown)


x110dc:
LGTM.

PR #46, opened 2018-08-23
FMD18: Thermometer; Campaign copy; Member benefits  

emilyyount:
Updates to support.texastribune.org/donate: 
- New background tile to match FMD branding 
- New rotating words 
- Added text for member benefit levels
- Including a couple more member quotes
- Thermometer based on Salesforce report. 

To-Do: 
- Link for "Spread The Word" button
- Replace SMD report JSON with FMD report JSON 
- Update text for /charge page. 

This has been browser tested.


static/js/src/entry/donate/Thermometer.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+  <div class="thermometer">
+    <div class="bar">
+      <div class="bar_inner bar_inner--goal"></div>
+      <div class="bar_inner bar_inner--actual" v-bind:style="{width: actualWidth + '%'}"></div>
```
AndrewGibson27: `v-bind:style` can be shortened to just `:style`. Also, I'd use template literals instead of string concatenation for `actualWidth + '%'`.

static/js/src/entry/donate/Thermometer.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+  <div class="thermometer">
```
AndrewGibson27: This will break if the AJAX call fails. So I'd have a `v-if` check in here [like Wall component does](https://github.com/texastribune/salesforce-stripe/blob/master/static/js/src/entry/circle/Wall.vue#L3) to prevent that.

static/js/src/entry/donate/Thermometer.vue
```diff
@@ -0,0 +1,61 @@
...
+      error: false,
+      data: [
+        { slug: 'actual', label: '0', value: 0 },
+        { slug: 'goal', label: '0', value: 0 },
+      ],
+    };
+  },
+  computed: {
+    actualWidth: function() {
+      return ((this.data[0].value / this.data[1].value) * 100);
```
AndrewGibson27: Should this logic ensure the width never exceeds 100%? Looks like it's not causing CSS overflow if it's above 100%, but it's possible some browsers might interpret it differently. Probably not -- I'm just being paranoid.

static/js/src/entry/donate/Thermometer.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+  <div class="thermometer">
+    <div class="bar">
+      <div class="bar_inner bar_inner--goal"></div>
+      <div class="bar_inner bar_inner--actual" v-bind:style="{width: actualWidth + '%'}"></div>
+    </div>
+    <div class="text">
+      <strong>{{ data[0].label }} new members</strong>
+      <br>
+      towards goal of {{ data[1].label }}.
```
AndrewGibson27: I can't believe I'm saying this, but isn't it toward instead of towards?

AndrewGibson27:
LGTM. Left some nitpicky comments. Two other things:
+ The language to the right of the thermometer reads like this: `$57,423.19 new members 
towards goal of 50,000.00.` That didn't quite make sense to me, but I could be dumb.
+ I'm wondering how the thermometer will play from an accessibility standpoint. It might be worth doing an `aria-hidden="true"` on the thermometer itself since screen readers probably won't be able to make sense of the percentage since it's just an inline style. And then make sure the text to the right of the thermometer is super explanatory. Or you could try to do some `aria-label` magic. Or something else I haven't thought of.

static/js/src/entry/donate/Thermometer.vue
emilyyount: Amanda would love this paranoia! I hope we exceed the goal too :) 

To account for that, in the CSS I set the div to overflow: hidden.

emilyyount:
Thanks, @AndrewGibson27! I think I fixed everything you mentioned. The branch is also updated with the real FMD report, so you won't see the strange dollar amounts in the text anymore.

PR #47, opened 2018-09-05
Better customer ID validation to prevent throttling

AndrewGibson27:
#### What's this PR do?
Puts the customer-retrieval portion of the Flask logic into a `try/except` block.

Twice now, a rogue IP address has slammed this app's `/charge` URL with bogus customer IDs. It has burned through our free Sentry quota and caused annoyance in `#tech-alerts`. We blocked the questionable IP, but it's possible someone else could attempt the same thing.

So far as I can tell, there isn't a way to syntactically validate the customer ID. But trying to fetch a non-existent customer by ID throws a custom Stripe error (`InvalidRequestError`), so catching that exception should do the trick. In that case, we simply send users to an error page.

+ `make interactive`
+ `yarn run dev`
+ On `master`, do the cURL command below. Confirm you get a Python error in the response -- something like `Could not validate nonexistent customer foobarbaz`
+ Now, on this branch, do the same cURL. There should be an HTML response with the error page. Somewhere in that, it should say `There was an issue saving your donation information.`
+ On this branch, try making a donations with different amounts and frequencies. Confirm you get the "Thank you" page.

Command:
```
curl -d "amount=1&campaign_id=foobar&customerId=foobarbaz&description=hi&first_name=mugnypqo&installment_period=yearly&installments=None&last_name=mugnypqo&openended_status=Open&pay_fees_value=True&reason=1&stripeEmail=sample@email.tst&zipcode=94102" -H "Content-Type: application/x-www-form-urlencoded" -X POST http://local.texastribune.org/charge
```

No.

NA.

No.

https://3.basecamp.com/3098728/buckets/736178/todos/1242027477

NA.

#### Next steps?
NA.

#### Smells?
Ideally we would just validate the syntax of a customer ID without having to catch an exception. But there doesn't seem to be a way to do that. Understandably so, too -- Stripe probably doesn't want to reveal its ID-generation secrets for security purposes.

#### TODOs:
NA.

Inline comment included.

#### Technical debt note
Same.

### DIFF
--- a/app.py
+++ b/app.py
@@ -244,8 +244,14 @@ def charge():
             gtm['event_value'] = request.form['amount']
             return render_template('charge.html',
                     amount=request.form['amount'], gtm=gtm, bundles=bundles)
-        except stripe.error.InvalidRequestError:
-            return render_template('error.html', message=error_message)
+        except stripe.error.InvalidRequestError as e:
+            body = e.json_body
+            err = body.get('error', {})
+            message = err.get('message', '')
+            if 'No such customer:' not in message:
+                raise e
+            else:
+                return render_template('error.html', message=error_message)
     else:
         print('Form validation errors: {}'.format(form.errors))
         print('Did not validate form of customer {} {} {}'.format(
--- a/app.py
+++ b/app.py
@@ -220,6 +220,7 @@ def create_customer():
 
 @app.route('/charge', methods=['POST'])
 def charge():
+    error_message = "There was an issue saving your donation information."
     gtm = {}
     form = DonateForm(request.form)
     pprint('Request: {}'.format(request))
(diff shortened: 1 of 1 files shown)


app.py
```diff
@@ -231,24 +232,29 @@ def charge():
...
+            print('Validated form of customer {} {} {}'.format(customer_email,
+                customer_first, customer_last))
+            if request.form['installment_period'] == 'None':
+                gtm['event_label'] = 'once'
+            else:
+                gtm['event_label'] = request.form['installment_period']
+            gtm['event_value'] = request.form['amount']
+            return render_template('charge.html',
+                    amount=request.form['amount'], gtm=gtm, bundles=bundles)
+        except stripe.error.InvalidRequestError:
```
x110dc: `InvalidRequestError` can be raised for other reasons. So I'd introspect the exception to see if it contains `No such customer:`. Otherwise, if that exception is raised for some other reason it'll be difficult to debug.

AndrewGibson27:
@x110dc Thanks for the review. I updated the logic to re-raise `InvalidRequestError`s if they aren't for bad customer IDs. I was able to test this by forcing by adding a bogus keyword argument to this [function call](https://github.com/texastribune/salesforce-stripe/blob/9496dd6c8fedfcfc4965e5e9c7ca08c574f27799/app.py#L236), which raises an `InvalidRequestError`. Upon cURLing to `/charge`, I got the Python error in the output (and in a Sentry email). When I removed that kwarg and cURL'ed again, I got the HTML error-page response.

x110dc:
I made one tiny change. LGTM.

PR #48, opened 2018-09-09
Charge new donations immediately

x110dc:
#### What's this PR do?

1. Charges new donations immediately instead of waiting for the batch job to run.
1. Makes some minor PEP8 changes.
1. Checks for duplicate transactions.

The batch job was designed to run once or a few times daily. But at certain times (like membership drives) they need to have more up-to-date numbers. So we run the batch job more frequently. But we can only configure it to run at most once an hour. Moving charging for single donations and the initial charge for new recurring donations to the time of submission means those numbers are always up-to-date without having to run the batch job frequently. 

1. Run the project as normal. 
1. Make some test transactions. Try regular, recurring, Circle and Blast. 
1. Make sure you see a message about charging the card and a successful call to the Stripe API for it.
1. Try new donations with and without `CHECK_FOR_DUPLICTES` set and check that it behaves as expected.

No, but existing ones have been modified.

N/A

This will make it take longer to process a transaction. But since it's done asynchronously in the background it won't mean any additional time for the user.

Off-sprint.

Let membership know that numbers will be up-to-date all the time.

#### Next steps?

Add the new `CHECK_FOR_DUPLICATES` to our test and production configuration.

#### Smells?

We don't really have a way to know if the user intended to donate twice in the same day. So it's possible that they will intentionally do so. But we'll get an alert and can contact the user and process it manually in those cases. And this will allow charges for different types in the same day: a Blast and a membership will go through. But for two recurring donations in the same day, the second will be stopped. But this behavior can be disabled with a new variable: `CHECK_FOR_DUPLICATES`.

Not yet.


x110dc:
https://github.com/texastribune/salesforce-stripe/pull/48/files#diff-ad1fae7000300b42844d2aee583f2602R389

You may not've seen it because the diff for `salesforce.py` is large so it was hidden. @AndrewGibson27

AndrewGibson27:
For some reason, the last three test charges I've done haven't caused the charge message or Stripe API call to show up in my terminal (there's also been nothing in the #stripe channel). That all worked when I initially reviewed this. I tried with `CHECK_FOR_DUPLICATES` set both to `True` and `False`.

I'm puzzled about why that could be. But it all worked for me initially.

x110dc:
I re-tried it just now and it worked. The only thing I can think of is that the celery worker wasn't running when you tried. But this can all wait until you get back!

PR #51, opened 2018-10-03
Better customer ID handling

AndrewGibson27:
@x110dc Sorry if my Python sucks.

#### What's this PR do?
+ De-ajaxes Stripe card validation and instead reloads a pre-filled form with a prominent error message if a `CardError` is raised in Python.
+ Improves server-side form validation with more granular expectations for a field's value.

So Stripe customer IDs are not visible on the client. Specifically:
+ No duplicate transactions from mistakenly resubmitting forms
+ No risk of someone gaining access to a customer ID and making a transaction on someone else's behalf

+ Build a fresh container with `make`
+ `make restart`
+ Fill out the form on `/donate`. Submit using either an error-raising test card number on [this list](https://stripe.com/docs/testing) or with a real card number. The page should reload with a red-boxed error atop the form. All fields besides card should be pre-filled with your previous input. The error should not disappear until you try to pay again.
+ Submit again, this time with a non-error-raising Stripe test card. The form should submit successfully, and you should see the thank-you page. Any referral or campaign IDs set as query params initially should have persisted through to this successful submit.
+ Repeat the last two steps on `/circleform`.
+ @x110dc Mind taking a glance at `forms.py` to ensure I didn't screw up anything glaringly obvious with the validation updates?

No. This app could use a front-end suite, but today is not that day.

Yes.

No.

Off-sprint.

NA.

#### Next steps?
NA.

#### Smells?
NA.

#### TODOs:
* [x] Verify this JSON is sanitized properly to prevent XSS. @x110dc I might request your help here since you're tackling this anyway.
* [ ] After deployment, I'll do my usual slew of $1 test donations, including with Apple and Google Pay.

It will be before I merge.

#### Technical debt note
Even. Adds complexity because of pre-filling form state. Removes some because we're ditching the `create-customer` endpoint.


app.py
```diff
@@ -112,21 +111,90 @@ def get_bundles(entry):
     return bundles
 
 
-@app.route("/donate")
+def do_charge_or_show_errors(**kwargs):
+    app.logger.debug("----Retrieving Stripe customer...")
+
+    try:
+        email = request.form["stripeEmail"]
```
x110dc: I'd move everything out of the `try` block except the call to Stripe. It makes it easier to reason about and more obvious that everything after that won't be executed if there's an exception.

forms.py
```diff
@@ -4,26 +4,37 @@
                             RadioField, StringField)
 from wtforms.fields.html5 import EmailField
 
-
-class DonateForm(FlaskForm):
+class BaseForm(FlaskForm):
```
x110dc: +1 for pulling the common elements into a base form.

app.py
```diff
@@ -112,21 +111,90 @@ def get_bundles(entry):
...
+        )
+
+    return render_template(template,
         bundles=bundles,
         key=app.config["STRIPE_KEYS"]["publishable_key"],
     )
 
-
-@app.route("/circleform")
+@app.route("/circleform", methods=["GET", "POST"])
```
x110dc: Maybe not in this PR, but I'd be in favor of making this consistent by dropping the `form` and renaming it to `circle` (and redirecting) like `donate`.

x110dc:
I don't feel strongly about this, but since the only error message that can be shown is credit card- related should the error message be above the credit card field?

app.py
AndrewGibson27: Done.

app.py
AndrewGibson27: I agree. I'll do that separately.

AndrewGibson27:
@x110dc Regarding the error-message location: There are a few technical challenges on my end to putting the error message above the card box on load (happy to detail them if you're interested). Also, IMO, it makes sense to be above the form when the page reloads as it's sort of an overarching reason why the whole submission failed. But I'm no UI expert and will confirm with Emily Y. that she's cool what I've done.

I fixed the escaping issue.

PR #53, opened 2018-10-12
Add reason to Salesforce when charge fails

x110dc:
#### What's this PR do?

Updates the opportunity with the reason when a card charge fails. And updates the status to "Closed Lost."

Right now we just keep retrying for three days and then give up. The opportunity is left in the `Pledged` state. Then someone has to either manually go through and mark them as `Closed Lost` or run a job to do so. And they don't know why the transaction failed. Sometimes that reason code is useful. These retries have not proved useful because the status rarely or never changes. So it gives membership more information about why a charge failed and it automates a process that would otherwise be manual.

- Enter a donation and use this card number: `4000000000000341`
- Then kill the server and run this in the container: `LOG_LEVEL=DEBUG python batch.py`
- Confirm that you see `Opportunity set to 'Closed Lost' with reason: Your card was declined.` in the output.

None added but some have been modified.

N/A

No.

https://3.basecamp.com/3098728/buckets/736178/todos/1089655685

Just let Anna know.

#### Next steps?

Start charging immediately so we don't have to run the batch job as frequently.

#### Smells?

I don't think so.

Will after merge.

#### Technical debt note

None.


AndrewGibson27:
LGTM.

PR #55, opened 2018-10-19
Get Sentry working with Celery 

x110dc:
#### What's this PR do?

Gets Sentry working with Celery.

Right now we're only getting Sentry notifications from the web process and since most of the processing happens with Celery we're missing out on errors generated there. Right now we're relying on alerts from logs but Sentry would be much better.

1. Set up your Sentry account if you haven't already done so. 
1. Set/change `ENABLE_SENTRY` in your `env-docker` file to `True`
1. Set/change `SENTRY_DSN` to the value provided by Sentry (or just ask me.)
1. Set `SENTRY_ENVIRONMENT` to `local` in `env-docker`.
1. At line 237 in `app.py` add `raise Exception('my funny phrase here')`
1. `make restart` and try to load `local.texastribune.org/donate`
1. Confirm that the exception was relayed to Sentry.
1. Move the line from the previous step to line 133. 
1. Try to make a donation. You should still get the "Thank You" screen but Sentry should have a new error. Confirm that.

No.

N/A

No.

https://3.basecamp.com/3098728/buckets/736178/todos/803864320

N/A


AndrewGibson27:
WFM

PR #56, opened 2018-10-19
Add security and rate limiting features

x110dc:
#### What's this PR do?

Adds several security-related headers and rate-limiting. Adds a Content Security Policy that's "report only" at the moment so we can confirm everything is working as expected before we start enforcing it.

It's good to be more secure. Especially when handling credit cards. 

1. set `REPORT_URI` in your `env-docker` to the value in Sentry (or ask me.) Start the server.
1. Visit a few pages. Confirm everything's working as intended. 
1. Comment out line 62 of `app.py`. Restart the server. Confirm that you get a warning on the browser console and an alert in Sentry.

No.

No.

Not likely. 

https://3.basecamp.com/3098728/buckets/736178/todos/1322629724

N/A

#### Next steps?

See if we can remove `data:`, `unsafe-inline`, and `unsafe-eval`.


AndrewGibson27:
Manual tests WFM. But I did get a `429 too many requests` from the Sentry endpoint in my browser console, though. Oddly, it looks like the Stripe script was the initiator. Can show you on my screen if it's helpful.

PR #59, opened 2018-10-30
Bus mem form and ui state - one dollar test

valasensio:
#### What's this PR do?
This PR adds the Business Membership to the Donations apps.
THIS PR IS FOR A $1 TEST VERSION which will be replaced once form/transaction processing on mobile (ApplePay, GooglePay) is validated in a production environment.

It creates a simple way for businesses to financially support the Texas Tribune.
The deploy timed to support of the Fall Membership Drive (<- not 100% certain about this).

* By entering valid and invalid data into the form
* By including valid and invalid query parameters in the URL to see if the form state changes per [spec](https://wiki.texastribune.org/display/TECH/Business+Membership+Page)

Engineering will notify Sponsorship and Art teams after the $1 test is completed, and the correct donation levels are updated and deployed.

There may be smells. First VueJS project for me, Andrew Gibson will review.
The technical debt consists of 2 remaining uses of Object.assign() which need to be replaced, since IE 11 dies not support Object.assign().

There are a handful of Basecamp issues across several teams. Deferring researching and aggregating those until we do the final PR for the busmem page.

* [ ] Tested manually on mobile? *(Not actually possible to test mobile payments in test environments)*
* [x] Checked for performance implications? *(This is a lightweight SPA; a single form that is unlikely to garner sufficient traffic to create any detectable server load)*
* [x] Checked for security implications? *(https; query param whitelisting; secure transaction processing tokenized using Stripe)*
* [x] Updated the documentation/wiki? *([ here](https://wiki.texastribune.org/display/TECH/Business+Membership+Page) )*

#### TODOs / next steps:

* [ ] *Once mobile transactions are validated, the code needs to be updated to replace the $1 donation level with the actual $2500 donation level*
* [ ] *Code and test replacements for Object.assign()*


valasensio:
circleci was down when I did the PR. I think circleci is back; I'm not sure if I should re-trigger those tests?

valasensio:
circleci passed. Great!

x110dc:
I'll leave the frontend bits to @AndrewGibson27 but all the functionality looks great from my view. LGTM.

valasensio:
@x110dc Thanks for checking it!

static/js/src/entry/business/BusinessForm.vue
```diff
@@ -0,0 +1,393 @@
...
+      <hidden
+        name="pay_fees_value"
+        store-module="businessForm"
+      />
+      <hidden
+        name="openended_status"
+        store-module="businessForm"
+      />
+      <hidden
+        name="customerId"
```
AndrewGibson27: Should be able to remove this.

static/js/src/entry/business/SelectList.vue
```diff
@@ -0,0 +1,85 @@
+<template>
+  <div
+    :class="classes"
+  >
+    <label>
```
AndrewGibson27: This label should be tied to the `<select>` with a `for` attribute, referencing the `id` of the `<select>`.

AndrewGibson27:
Manual testing all WFM.

static/js/src/entry/business/SelectList.vue
valasensio: Thank you! Will address before production deploy.

static/js/src/entry/business/BusinessForm.vue
valasensio: Okay,. Makes sense, not being used. Will clean it out.

static/js/src/entry/business/SelectList.vue
valasensio: Question: That particular label is the "STATE" label above the field where the state selection shows. If I 'for' it it will show 'STATE' 51 times above the form field. Am I misunderstanding your intention with this comment?

AndrewGibson27: Different `for` -- not `v-for` but rather [this kind](https://stackoverflow.com/questions/698001/whats-the-for-for-in-a-label-tag). In other words, the `<label>` element should have an attribute `for="foo"`, and the `<select>` element should have `id="foo"`. That way, screen readers will be able to tie the text in the label to the actual form input.

valasensio: Ok. Got it. Thank you for the pointer to the example for clarification. 
Sincerely, 
Your humble Vue Nue-bie

AndrewGibson27: Vuebie!

valasensio: That works!

PR #60, opened 2018-10-31
Initial production version of the Business Membership page.

valasensio:
#### What's this PR do?
This PR adds the live version (non-$1 test) Business Membership to the Donations apps.

It creates a simple way for businesses to financially support the Texas Tribune. It is targeted to a different group than the Circle Membership or the regular Donations apps.

* By entering valid and invalid data into the form
* By including valid and invalid query parameters in the URL to see if the form state changes per [spec](https://wiki.texastribune.org/display/TECH/Business+Membership+Page)

Engineering will notify Sponsorship and Art teams when page is deployed to production.

Designed the code with the objective that when the Sponsorship and Art teams want to make changes, which is inevitable, they can be made quickly by changing constants rather than rewriting code. 

The query parameters definitions have to be mapped between UI form state and POST form state with one particular parameter triggering cascading state changes. This is a little confusing, but I've tried to document it to clarify. In several cases, the string names displayed in the UI differ from the POST form strings required on the server side for a given state, so those are also mapped across states as well.

The main Basecamp project is [here](https://3.basecamp.com/3098728/projects/9070027).

* [ ] Added automated tests? *( TBD)*
* [x] Tested manually on mobile? *()*
* [x] Checked for performance implications? *(This is a lightweight SPA; a single form that is unlikely to garner sufficient traffic to create any detectable server load)*
* [x] Checked for security implications? *(https; query param whitelisting; secure transaction processing tokenized using Stripe; form fields are validated and length limited; no alien scripts will be executed from the form input data)*
* [x] Updated the documentation/wiki? *([ here](https://wiki.texastribune.org/display/TECH/Business+Membership+Page) )*

#### TODOs / next steps:

None noted


valasensio:
Thanks Andrew! All good points above. Noted. You've been a huge help! Hoping the Sponsorship team sees some action on it.

valasensio:
@AndrewGibson27 Following up on query param filtering: It can be a fairly complex system; there are probably libraries for it. But I picked a simplified middle path based on Daniel's mentioning at a BusMem meeting that he filters the parameters (against SQL injection attack etc) on the back end. 

That leaves the front-end filtering to prioritize short circuiting query parameter overloading attacks by recognizing and not processing obviously bogus data as early as possible. 

The filtering is along 2 vectors: 

**Whitelisting to ignore invalid param keywords**
* Params not on the whitelist are ignored
* Valid params are recognized and processed (sent to the back end)
* Both can exist in the same query
* The special parameter key ```?Installments=once``` is validated and only the value 'once' causes UI and form state update. An invalid query parameter ``?Installments=gah``` is ignored and default settings are used.

**Bandwidth limiting**
* If the number of query parameters exceeds the given threshold, the query params are not processed and defaults are loaded....nice try, villain. 
* Each valid query param value is limited to a threshold number of characters. It's concatenated to the number of characters limit. 

More extensive query filtering is definitely possible. Regexing is always good for numeric fields. And, the obvious solution for string values is to add a validation lookup table for all the keywords and boot any values that do not comply--see my above comment about ```installaments```. That would be an additional task. I'm not sure there's a lot of bang for that buck though. But, I don't have enough information to say either way.

**Form security is a thing**
As none of the other forms have query filtering and there hasn't been an issue, I concluded it's not been a problem. But I added this bit of filter because query parameters are a notorious security hole (Rails is infamous for it) so whitelisting parameters is basic and important level of form security if you're ingesting query parameters.

valasensio:
I'll add the above to the wiki

PR #61, opened 2018-11-07
Enforce reason character limit client side

AndrewGibson27:
#### What's this PR do?
Gives you an error message if you try to enter an "encouraged by" longer than 255 characters

To line up with the server-side validation.

On both `/circleform` and `/donate`, try entering an encouraged-by longer than 255 characters. Confirm you can't submit the form until you shorten it.

NA.

NA.

https://3.basecamp.com/3098728/buckets/736178/todos/1403846031

* [ ] Added automated tests? *(NA)*
* [ ] Checked for security implications? *(NA)*
* [ ] Updated the documentation/wiki? *(NA)*

#### TODOs / next steps:
* [ ] There are character limits to enforce for the business-membership form. But I'll leave those to @valasensio because I don't know the status of that page's launch.

### DIFF
--- a/static/js/src/entry/circle/CircleForm.vue
+++ b/static/js/src/entry/circle/CircleForm.vue
@@ -70,10 +70,13 @@
         <div class="col_6 grid_separator">
           <text-input
             :required="false"
+            :show-error="showManualErrors || showNativeErrors"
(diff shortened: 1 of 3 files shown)


x110dc:
LGTM.

PR #64, opened 2018-11-16
Save (and update) more credit card details to Salesforce

x110dc:
#### What's this PR do?

It adds expiry date, card brand and last 4 digits to the list of card details we save in Salesforce when the donation is made. If that data changes on the Stripe side after that it will also update it in Salesforce.

It allows us to send reminders to people about renewals that include details about what card is being charged. It may also allow us to pre-emptively ask members to update their credit card if it will expire before their renewal.

It's hard to test the server portion locally so I've put it up on test. But you'll need to run some code locally. 

- `make`
-  Add this code to a temporary file (say `tmp.py`):

```
import stripe
customer = stripe.Customer.retrieve("cus_DovRfzO8debJZl")
card = customer.sources.retrieve(customer.sources.data[0].id)
card.exp_month = 4
card.exp_year = 2022
card.save()
```
- run that code: `python3 tmp.py`
- look in the logs for the test server to confirm that a request came in and was processed without error
- make a new donation on the test server 
- confirm in the logs that everything processed without error

Let membership know.

No.

https://3.basecamp.com/3098728/buckets/736178/todos/1291799118

* [x] Added automated tests? *(one was updated)*
* [ ] Tested manually on mobile? *(N/A)*
* [ ] Checked for performance implications? *(N/A)*
* [x] Checked for security implications? *(The webhook signature is verified)*
* [ ] Updated the documentation/wiki? *(will after merge)*


AndrewGibson27:
WFM

PR #81, opened 2018-12-05
Add pre-commit JS formatting; integrate ESLint into CI

AndrewGibson27:
#### What's this PR do?
+ Adds JS code formatting via Prettier, which is done automatically pre-commit
+ Makes a CircleCI build fail if linting via ESLint fails

+ Time spent in code review talking about style is time wasted. Prettier takes care of JS formatting without anyone having to even think about it.
+ JavaScript is changing at an extraordinary rate, meaning it's easy to do things the old/ES5 way simply because you don't know a better way exists. By enforcing ESLint rules, we'll always be up-to-date on modern JS syntax, especially when it comes to Vue.js.

+ Install the pre-commit package: `brew install pre-commit`
+ Ensure you have a relatively up-to-date version of Node on your machine (> 8 ideally).
+ Do `pre-commit install` at the root of this directory.
+ Create a branch off this one (`git checkout -b foo-branch js-ci`)
+ Remove the semicolon from [this line](https://github.com/texastribune/donations/blob/master/static/js/src/utils/addNumberCommas.js#L2).
+ Try committing that change and confirm you get a failure message. The JS file should've been corrected to now have a semicolon. Do `git status` and verify that the semicolon addition didn't get automatically committed and is waiting to be added to the staging area. (I'm opting for this approach after reading [this](https://github.com/pre-commit/pre-commit/issues/747#issuecomment-386782080)).

NA.

Our CircleCI builds will take 15-20 seconds longer now.

https://3.basecamp.com/3098728/buckets/736178/todos/1457497379

* [ ] Added automated tests? *(NA)*
* [ ] Checked for security implications? *(NA)*
* [x] Updated the documentation/wiki?


x110dc:
You may want to add something to `STYLEGUIDE.md`.

AndrewGibson27:
@x110dc Done.

x110dc:
You may want to note (in the README? in STYLEGUIDE? both?) that running `yarn` will add to or modify files in your `.git/hooks` directory. It was a surprise to me that something would do this without requesting permission or at least notifying me that it was doing this. 

Also, this won't work if you already have a `pre-commit` hook in place. For example, if you're using this for to run/manage your pre-commit hooks:

https://pre-commit.com/

Then to get the same behavior you'll need to do something like this:

https://prettier.io/docs/en/precommit.html#option-3-pre-commit-https-githubcom-pre-commit-pre-commit

Or this:

https://github.com/awebdeveloper/pre-commit-prettier

x110dc:
LGTM with the documentation notes listed above.

AndrewGibson27:
@x110dc Thanks for reviewing. I opted to go with the `pre-commit` package instead of `husky` because I think it'll play nicer with multi-language repos in the long run. If you don't mind giving another look, I updated the manual steps.

Feel free, of course, to uninstall `pre-commit` and clear out all the hooks it generates when you're done here.

x110dc:
It looks great! Thanks.

PR #99, opened 2019-01-16
CSS reorg round 2

ashley-hebler:
#### What's this PR do?

- Removes duplicate element resets and restated classes
- Removes explicit CSS property assignments when they can naturally inherited (ie font family, color, etc)
- Moves all membership drive css into new components:  `6-components/thermometer` and `6-components/smd`
- Creates a `.splash` component with updated BEM-ifying and repurposes that for both the /donate and /blastform headers
- Remove use of `@extend` for fonts in favor of `@mixin`
- Brings all remaining views into the new CSS ecosystem

We're trimming fat in all the outputted files. This will put us in a good place to start thinking of our templates as modular UI elements

Visit every possible view of the /donations app and check for any style inconsistencies. Nothing should be wildly different from what's live.

N/A

We're left with a bunch of files; mass deletion round of those to come.

* [ ] Added automated tests? *( N/A )*
* [ ] Tested manually on mobile? *( will after PR deploy)*
* [ ] Checked for security implications? *( N/A )*
* [ ] Updated the documentation/wiki? *( N/A )*

#### TODOs / next steps:

* [x] Get business and circle on the same CSS base (remove dup CSS)
* [x] Integrate blastform (/blastform) into new structure - create components where needed
* [x] Integrate homepage (/donate) into new structure - create components where needed
* [x] Replace silent classes with mixins
* [ ] Turn one-off page styles into reusable components (example circle-wall)
* [ ] Delete legacy CSS*
* [ ] Remove opinionated sizing on heading tags
* [ ] Convert class names to use UI namespacing c-component o-object
* [ ] Scope out adding our own spin to ITCSS with a typography folder (maybe after objects/before components)
* [ ] Clean up objects modules. Remove specifically-styled files like borders and move those into to components


ashley-hebler:
Another monster PR for ya @AndrewGibson27. Sorry for that! Let me know if you want to discuss and take your time on getting around to this review.

static/sass/6-components/_error-message.scss
```diff
@@ -0,0 +1,33 @@
+.error-message {
```
AndrewGibson27: I'd argue this is more a typography style than a component. But I know dealing with type styles is slated for later, so no need to address now.

ashley-hebler: Fair point to bring up. It’s basically just repeating our link styles so it’s not super necessary as a stand-alone anything. I’ll remove it, and apply that look another way. Thanks!

ashley-hebler: screen shot 2019-01-17 at 10 12 10 am

So if I ditch that all together and just slap your fav class `.prose` on the parent, it replicates the desired look, but with a serif. How does that sound in the interim until we have our typography styles refined, @AndrewGibson27?

AndrewGibson27: WFM!

AndrewGibson27:
So there are two pages that are looking a lil' off (they're kinda buried in the app and hard to find):
- `/submit-blast` (the page you get when you successfully submit a Blast payment; uses the `blast-charge.html` template)
- `/blast-promo` (a temporary page for a promo we're running on Blast subscriptions; will die soon)

I defer to you on how to style the former; it could probably use a lot of stuff from the regular thank-you page. As for the latter, it's virtually the same page as `/blastform`, so I'm sure it can use the same isolated, icky styles.
--
But aside from those, this is wonderful! More delightful changes. Cheers to the component world and organizing things in a way that actually makes sense. All other pages look good and work for me.

This is somewhat contrary to one of my comments on your last PR. But I'm increasingly thinking maybe we don't need `5-objects` at all. One of the articles you linked to mentioned it's acceptable to drop that bucket: "I personally prefer to skip it entirely to avoid confused mental mapping. That avoids me to stop and think: 'Is that an object or a component?', and to remain focused on the real problem."

I'm starting to agree. First, I'd hate for us to waste lots of time in code review saying, "Hey, _actually_, you put a single measly border on that unordered list and thus it's a component, not object. Go move that file now!" Second, doesn't it already kind of feel like most everything is a component -- just looking at the number of files in `6-components` compared to `5-objects`? The latter will look even more bare if:
+ We move `_typography.scss`, and `_links.scss` `_fontawesome.scss` into their own "typography" sub-directory, as we've discussed.
+ Move `_lines.scss` to `6-components`, where I feel it belongs anyway since it's very much color-dependent.
So that leaves us with just `_grid` and `_images`. The state of our grid system is TBD, so eh. And I think `_images` could just as well be a component. I'll end my rant there. But having slept on it, I do think our lives could be simpler if we drop objects.

I wanted to note a few more legacy style smells. Given these also exist in `texastribune`, I'd say there's no need for you to address them as part of your refactoring of this app. Instead, we can tackle them as a whole when we get to some kind of cross-repo design system. But, just for the record:
+ I hate our links, mostly because you can't apply those classes directly to anchor tag and instead have to put it on a parent element. And they use `extend` like crazy.
+ I hate our buttons, mostly because of the crazy-long mixin that builds them.
+ Our grid system should definitely be replaced with CSS Grid at some point. But there are still discussions about IE 11 support and such.

Lastly, we should figure out what to do with the nav. IMO, having the same nav as the main site doesn't even make sense in this app. How should we expect someone giving us money to know what "our picks" is, for example. Also, the user portal will go in this repo, and that will bring in some design considerations of its own. So probably worth just delaying those decisions. But curious what you think.

ashley-hebler:
@AndrewGibson27,

I def didn't know about those pages so I'm glad you checked. Should I add this to the readme?


------
## Various views of this app

### Pages
- [/donate](http://local.texastribune.org/donate)
- [/business](http://local.texastribune.org/business)
- [/circle](http://local.texastribune.org/circle)
- [/blastform](http://local.texastribune.org/blastform)
- [/blast-promo](http://local.texastribune.org/blast-promo)
- [/error](http://local.texastribune.org/error)

### Success/Error states
- 404
    - Template is `error.html`
    - This displays when the app encounters an unknown slug
- Successful donation submission
    - Template is `charge.html`
    - This displays for forms on `/donate`, `/business`, and `/circle`
- Successful subscription submission
    - Template is `blast-charge.html`
    - This displays for forms on `/blastform`, `/blast-promo`
-----

I don't think you could have described my own hesitations/confusion around the objects designation more perfectly. I ditched that so the new dir tree will look like this:

    .sass
    ├── settings/       # variables
    ├── tools/          # mixins and functions
    ├── generic/        # resets
    ├── elements/       # html tags
    ├── typography/     # classed, type-driven styles
    ├── components/     # UI custom styles
    ├── utilities/      # helpers and trumps
    ├── donate-update.scss
    ├── charge-update.scss
    ├── old-update.scss
    ├── circle-update.scss
    └── business-update.scss


- I changed overrides to utilities, which would allows us to use something like `.u-text-center`. It felt like a good home for grid as well while we soul search on that.
- Agree on all your links, buttons, and grid gripes - lots of improvements we could hash out for those
- The navbar actually feels right to me. (Whoa, we disagreed!) I know the donations form is part of a separate app only because of the codebase separation. As user if I read a story and decided to donate, a change in navbar would feel disjointed and perhaps off-putting to be pulled into a new portal where I'm now paying. When I think of myself less as a passer-by and more of a logged-in user, that lens changes and a more admin-driven bar would make sense to me. In that respect I see your point and that's good context to consider moving forward.

AndrewGibson27:
All of your comments look good to me. Feel free to update the README, though it's also possible to just crack open `app.py` and browse all the routes there.

As far as grid, I think putting it in utilities is good for now. Eventually, as I've probably said about 90 times, I'm a big proponent of designating all layout classes with a `l-`. I like that because it says to me, "Any component nested inside me (anything with a `c-`) will just fill the space I provide to it." But no need to worry about that now until we agree on a grid future.

PR #101, opened 2019-01-18
Create css components to match vue components

ashley-hebler:
#### What's this PR do?
Moves the form buckets and wall vue components into corresponding SCSS component files

A more DRY approach; ties into our componentized system of styling

A quick look at the [business](http://local.texastribune.org/business) and [cicle](http://local.texastribune.org/circle) pages.

N/A

I don't think so

https://3.basecamp.com/3098728/buckets/736178/todos/1516990374

* [x] Tested manually on mobile? *( will do this with PR link )*

#### TODOs / next steps:

* [x] Get business and circle on the same CSS base (remove dup CSS)
* [x] Integrate blastform (/blastform) into new structure - create components where needed
* [x] Integrate homepage (/donate) into new structure - create components where needed
* [x] Replace silent classes with mixins
* [x] Clean up objects modules. Remove specifically-styled files like borders and move those into to components
* [x] Scope out adding our own spin to ITCSS with a typography folder (maybe after objects/before components)
* [x] **Turn one-off page styles into reusable components (example circle-wall)**
* [ ] Delete legacy CSS
* [ ] Remove opinionated sizing on heading tags
* [ ] Convert class names to use UI namespacing c-component o-object


static/js/src/entry/business/Choices.vue
```diff
@@ -1,18 +1,16 @@
 <template>
```
AndrewGibson27: Nitpicky, but should we remain the two `Choices.vue` files `FormBucket.vue` or something like that so they line up more neatly with the class names?

templates/circle-form.html
```diff
@@ -119,8 +119,8 @@ <h4 class="grid_separator circle-benefits__header">Questions?</h4>
         </div>
       </section>
 
-      <section class="circle-wall grid_separator--l">
-        <figure class="circle-wall__logo grid_separator--l">
+      <section class="wall grid_separator--l">
```
AndrewGibson27: The wall is half in HTML land, half in Vue land (my fault). Would it make sense to just have the entire thing in Vue, maybe allowing the component to accept a logo prop?

ashley-hebler: yeah that bugged me too, but I didn't want to get too wild changing things. If you're cool with that, I'd prefer keeping it all in vue

static/js/src/entry/business/Choices.vue
ashley-hebler: for sure, always a fan of matching up like that. I'll get that updated. Thanks!

PR #105, opened 2019-01-23
Add icon build system and implementation of svgs

ashley-hebler:
#### What's this PR do?
- Adds a build task for SVG sprite generation
- Inlines said sprites
- Swaps out the font icon instances with `<svg><use>...`

- No longer tied to a lib for our icons
- 1 fewer http request

- Mostly present in the footer so check that for display of icons

N/A

Adds more packages, which could increase build time. 

https://3.basecamp.com/3098728/buckets/736178/todos/1516993849

* [ ] Tested manually on mobile? *( Will test after PR branch builds)*
* [x] Checked for performance implications? *(Ran some tests between diff methods, mostly negligible but reduction in requests and file size is always a win  )*
* [x] Checked for security implications? *(SVGs have security risks when it comes to users uploading to your app but we're not doing that. I also ran a `yarn audit` and these packages didn't add warnings )*
* [ ] Updated the documentation/wiki? *(Will do so after this is finalized )*

#### TODOs / next steps:

* [ ] *your TODO here*


package.json
```diff
@@ -24,14 +25,19 @@
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-preset-env": "^1.6.1",
     "css-loader": "^0.28.11",
+    "directory-tree": "^2.2.1",
     "es6-promise": "^4.2.4",
     "file-loader": "^1.1.11",
+    "fs": "^0.0.1-security",
```
AndrewGibson27: Unless I'm missing something, you shouldn't need to install `fs` explicitly as it comes with Node out of the box.

package.json
```diff
@@ -24,14 +25,19 @@
...
     "mini-css-extract-plugin": "^0.4.0",
+    "mkdirp": "^0.5.1",
     "node-sass": "^4.9.0",
     "optimize-css-assets-webpack-plugin": "^5.0.1",
+    "path": "^0.12.7",
```
AndrewGibson27: Same thing with `path`.

PR #175, opened 2019-04-29
Make BusMem query params work like Circle ones

AndrewGibson27:
#### What's this PR do?
+ Makes `?level=<level-name>` work on the BusMem page, just like it does on Circle
+ Consolidates the two `FormBuckets.vue` components into one

I'm trying to get this app's front end as clean as possible before it gets a major workout during UMP development. That means consolidating duplicate components and standardizing as much as possible. 

This will also allow our revenue people to send potential customers the BusMem page with a pre-selected level. I also think it makes sense to just use a single `level=` parameter as opposed to separate ones for installment period, amount, etc., given that subsidiary info can be derived from a particular level constant.

Cheers to Val for getting this page off to a wonderful start!

+ If you're not testing on the Heroku branch, do a `yarn` then a `make restart`
+ Visit `/business?level=<bigTexYearly|bigTexMonthly|loneStarYearly|loneStarMonthly|hatsOffYearly|hatsOffMonthly>` and ensure the proper bucket radio button is selected on page load. Make a few test donations and ensure the info comes thru properly in #stripe.
+ Make a few test donations on `/circle` just for good measure. Ensure the info comes thru properly in #stripe.

### How should this change be communicated to end users?
I'll let April know and then she can decide.

No.

https://3.basecamp.com/3098728/buckets/736178/todos/1493933269

* [ ] Added automated tests? *(someday)*
* [ ] Checked for security implications? *(we're now only processing a single query param, and only processing it if it contains a valid `level` value, so I feel good about things)*
* [ ] Updated the documentation/wiki? *(NA)*


x110dc:
I've been thinking about something similar. Let me know if it should go into this PR or not. Now that we have different endpoints for the different types (Biz, Circle, etc) the only parameter that needs to be passed to the backend is the installment period. Installments and open-ended status don't need to be sent to the backend at all. Hopefully, that will simplify parts of the frontend. What do you think?

AndrewGibson27:
@x110dc I think it's a great idea. I'd prefer it not go in this PR just for scope/number of files changed purposes, but I'll be happy to make the needed front-end changes to accommodate.

AndrewGibson27:
@x110dc I'd like to get this PR merged before we do [this](https://3.basecamp.com/3098728/buckets/10357222/todos/1787329290). I requested a review; please look at your convenience -- no big rush.

PR #186, opened 2019-05-08
Integrate design system build tasks

ashley-hebler:
#### What's this PR do?

Compiles SCSS into a `static/css` and builds the base SVG sprite for later icon usage.

Shifts our CSS/iconography config to one shared build system. For example, when we want to make changes to autoprefixer or any other post CSS setting we do it in [the toolbox](https://github.com/texastribune/ds-toolbox) and the styles in this repo will inherit those upgrades. In the future our other repos will also inherit them.

**TL;DR one source of truth for the front-end (sans JS) build** 

Run a fresh `yarn`

Run `yarn run dev`
 - Check for `Success! Watching styles...` and make sure webpack is still doin' its thing.
 - Break the SCSS by commenting out an important variable like `$color-yellow-tribune` or something. Make sure it barks at you. Webpack will too.

Question that came up in this: _Would it be correct to say `yarn run dev` does *not* start the app server anymore? And for this we now rely on `make restart`?_

Run `yarn run postinstall`
 - Check for `Success!` and make sure webpack is still doin' its thing.
 - Break the SCSS by commenting out an important variable like `$color-yellow-tribune` or something. Make sure it barks at you. It won't get to webpack bc it exits with `process.exit(1);`.

 
N/A

We'll want to switch the devDependency from `https://github.com/texastribune/ds-toolbox.git#dependencies` to `https://github.com/texastribune/ds-toolbox.git#master` once this is approved.

No specific task, but in the UMP world, this tees us up for UMP using the design system 🎉 

* [ ] Updated the documentation/wiki? *(will do once we hammer out the details )*

#### TODOs / next steps:

* [ ] *Hook up the styles to the new location*
* [ ] *Start using the style base from ds-toolbox (lonnng term, but UMP could!)*
* [ ] *Remove webpacks CSS importing? Maybe this deserves a discussion.*
* [ ] *Start using SVG sprite by overhauling font-awesome like we did in our main repo*
* [ ] *Update docs*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ celerybeat.pid
 yarn-error.log
 .pytest_cache
 .vscode
+# SVG sprite
+/templates/includes/base_icons.html
--- a/.gitignore
+++ b/.gitignore
@@ -3,12 +3,11 @@ env-docker
 .env
 *.pyc
 /static/css/
-/static/build/
-/static/prod/*.js
-/static/prod/*.css
-/static/prod/*.jpg
-/static/prod/*.json
-/static/prod/*.png
+/static/build/*.js
+/static/build/*.css
+/static/build/*.jpg
+/static/build/*.json
+/static/build/*.png
 node_modules/
 *.DS_Store
 .cache
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,6 +25,7 @@ ARG AUTH0_CLIENT_ID=k66bypGeKUCCLHGrQt5BQHJWTltZ9OzQ
 
 COPY static /app/static
 COPY webpack /app/webpack
+COPY config /app/config
 COPY package.json /app/
 COPY yarn.lock /app/
 COPY .babelrc.js /app/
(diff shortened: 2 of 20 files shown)


AndrewGibson27:
OK @ashley-hebler, I did some updates to get this working with Webpack, Heroku and long-term caching. My changes:
+ I changed to where CSS files get built. Heroku is unable to create directories during a deploy, so I used one with a `.gitkeep` in it.
+ I moved some stuff from `devDependencies` to `dependencies` because Heroku only installs packages from the latter during deploys.
+ I made it so Webpack processes CSS after it's built by your Node scripts. This is because it intelligently adds a cache-busting hash to stylesheets. _I don't like this solution long-term because processing CSS twice feels weird. But there's no way to bust the cache with the existing Node tasks, and not doing so will cause breakage for repeat visitors. We should add hashing to the DS to-do list._

Because of that last bullet point, when you run `yarn run dev`, there's initially going to be lots of red in the terminal because Webpack will be looking for a stylesheet that isn't built yet. But once the Node task finishes, Webpack will notice the change and then update.

To answer your question about commands:
+ `make restart` is the easiest way to run the app -- it'll start Flask, Webpack, Node and Celery
+ If you don't want to actually process test transactions on the back end, you can do `yarn run dev & python3 app.py` which will run Flask, Webpack and Node

PR #192, opened 2019-05-13
Remove CSS from webpack bundle

ashley-hebler:
#### What's this PR do?

Adds cache busting of CSS files outside of Webpack

Currently, we are running our CSS through a custom task (for scss compiling and optimizing) and Webpack (for cache busting). To cut down on running our CSS through 2 processes, we're making all the logic happen (compiling, optimizing, and cache busting) in the custom task.

TL;DR
Old way: Custom task + Webpack = final CSS
New way: Custom task = final CSS

_Test the build_
`yarn` then `make restart`
You should not see any red Webpack warnings nor should you see any CSS files referenced in the bundle summary.

Screen Shot 2019-05-13 at 12 15 57 PM

_Test the output_
Everything look normal here?
[http://local.texastribune.org/donate](http://local.texastribune.org/donate)
[http://local.texastribune.org/business](http://local.texastribune.org/business)
[http://local.texastribune.org/circle](http://local.texastribune.org/circle)
[http://local.texastribune.org/blastform](http://local.texastribune.org/blastform)
[http://local.texastribune.org/blast-promo](http://local.texastribune.org/blast-promo)
[http://local.texastribune.org/404example](http://local.texastribune.org/blast-404example)

Just a tricky note that this should be accompanied with a [ds-toolbox release](https://github.com/texastribune/ds-toolbox/pull/17).

Not directly, but should be helpful for future CSS iterations for UMP and otherwise.

* [ ] Updated the documentation/wiki? *(our CSS instructions in general probably need some updating. Adding as a todo)*

#### TODOs / next steps:

* [ ] *Add updated process to our Front-end wiki*


app.py
```diff
@@ -237,18 +237,28 @@ def get_bundles(entry):
...
+    root_dir = os.path.dirname(os.getcwd())
+    build_dir = os.path.join("static", "build")
+    asset_path = "/static/build/"
+    manifest_path = os.path.join(build_dir, "styles.json")
+    with open(manifest_path) as manifest:
+        assets = json.load(manifest)
+    css = assets[entry]
+    css_path = asset_path + css
+    return css_path
+
```
ashley-hebler: @mailbackwards, I requested your review for all the python edits I made. Does this all check out on your end?
Generally speaking, I'm taking some JSON like this:
```
{
  "business": "business.1557767697331.css",
  "charge": "charge.1557767697562.css",
  "circle": "circle.1557767697751.css",
  "donate": "donate.1557767697982.css",
  "old": "old.1557767698239.css"
}
```
and passing that context to each route. Like on [this line](https://github.com/texastribune/donations/pull/192/files#diff-3f41e546893dc64b71aaacad12cad815R407) with `css = get_styles("donate")`

mailbackwards: This LGTM-- it could be a little bit brittle, for instance if someone puts a typo in one of the arguments then it will error out the whole page, if you'd rather it error more "silently" and show the page w/o CSS, you might want to make it more forgiving by checking for the key in the dict first, and/or type-checking before concatenating them. But if we did that we'd also want to manually throw an error to Sentry so that we'd know something was up, so it just adds lotsa complexity for a very unlikely case...so I think it's fine for it to error loudly.

Conceptually/longer-term, I could see drawing out `get_bundles` and `get_styles` into a common underlying function since they have plenty of overlap in usage+details, but probably overkill and outta scope here. WFM!

ashley-hebler: @mailbackwards thank you and yes that's true on the "entry" param being a bit brittle. I was starting to feel like separating `get_bundles` and `get_styles` felt kinda repetitive. I'll see if I can add the `get_styles` extraction to `get_bundles`. Then when we fix the how that errors, we do it in one place. Good feedback, thanks!

ashley-hebler:
@AndrewGibson27, ooh that would be nice. So basically diff the compiled CSS and only update the revision number if it changes right? That makes sense so that we're not constantly emptying the cache if we don't need to. If I'm not over simplifying, I might be able to work that in. Let me experiment a bit in ds-toolbox

app.py
```diff
@@ -449,12 +474,14 @@ def the_blast_promo_form():
...
         key=app.config["STRIPE_KEYS"]["publishable_key"],
         bundles=bundles,
+        css=css,
     )
 
 
 @app.route("/submit-blast-promo", methods=["POST"])
 def submit_blast_promo():
     bundles = get_bundles("old")
+    css = get_css("old")
```
tjb346: Should this be get_styles? Just comparing to the other similar calls, and noticed this is the only place get_css is used.

ashley-hebler: yes, omg good catch! I'm working on `get_bundles` doing all the work though. So I'll remove in a future commit.

app.py
```diff
@@ -237,15 +237,17 @@ def get_bundles(entry):
...
     with open(manifest_path) as manifest:
         assets = json.load(manifest)
     entrypoint = assets["entrypoints"][entry]
     for bundle in entrypoint["js"]:
         bundles["js"].append(asset_path + bundle)
-    for bundle in entrypoint["css"]:
-        bundles["css"].append(asset_path + bundle)
+    with open(css_manifest_path) as manifest:
+        css_assets = json.load(manifest)
+    bundles["css"] = asset_path + css_assets[entry]
```
ashley-hebler: @mailbackwards I ditched get_styles in favor of just _bundling_ with get_bundles. Is that fair and did I break any bad rules in doing so?

mailbackwards: That seems really nice and clean to me, but I will defer to more expert bundlers in terms of whether this breaks any rules or makes things less flexible. I didn't realize you could [access dict keys like this](https://github.com/texastribune/donations/blob/746fe9795cbc0fdf3703f9048d5672f5a9971542/templates/new_layout.html#L26) in Flask, that's neat...don't think you can do that in Django templates.

ashley-hebler: cc: @AndrewGibson27, master of bundlers :)
I think one _could_ argue that now our CSS file names are a little more rigid, but I'm ok with that.
Oh haha well, maybe it's an antipattern? I think I thought that'd be alright to do since I was replacing something like this before...
```django
{% block stylesheets %}
  {% for stylesheet in bundles['css'] %}
    <link rel="stylesheet" type="text/css" href="{{ stylesheet }}">
  {% endfor %}
{% endblock %}
```
Would it be less surprising to see it as `{{bundles.css}}` or is it just that you're used to seeing the context give the value straight up without having to access it via key?

mailbackwards: I don't know of any problem with it, especially if it was already looking like that...I think it's true that you can do that in Django too with dot notation. @tjb346 might know more

package.json
```diff
@@ -27,19 +27,16 @@
     "@babel/runtime": "^7.4.3",
     "axios": "^0.18.0",
     "babel-loader": "^8.0.0",
-    "css-loader": "^1.0.0",
+    "css-loader": "^2.1.1",
```
ashley-hebler: @AndrewGibson27 I ended up adding this back since vue-loader was barking about it being a missing dependency. I wanted to play it safe, but if you want to remove later, go for it!

PR #200, opened 2019-05-20
Portal CSS

AndrewGibson27:
#### What's this PR do?
Styles the UMP pages à la the design system and some local styles that mirror the directory structure of the DS.

From a JS perspective, everything is still hard-coded. This isn't linked to an API yet.

So everything looks pretty.

+ `make`
+ `make restart`
+ Go to `/user-portal` and sign in with a test Auth0 non-staff account.
+ Click around the pages linked from the side nav. Make sure everything looks pretty and not broken and in line with these [mockups](https://3.basecamp.com/3098728/buckets/10357222/vaults/1746618729).

NA right now.

@ashley-hebler Please let me know if there are places I'm not using our shared classes properly or am duplicating stuff that's already accounted for in our utilities. I've also got a running list of stuff I think we should eventually add to the DS based on my experience styling this app.

As far as repo-specific styles go, I created `static/sass/ds`, which contains a mirror of the DS' directory structure. That's where my UMP styles are. I know the top level `static/sass` also mirrors that structure, but those partials of course predate the DS and thus I wanted to explicitly distinguish between pre- and post-DS worlds.

https://3.basecamp.com/3098728/buckets/10357222/todos/1720708167

* [ ] Added automated tests? *(later)*
* [ ] Tested manually on mobile? *(browser testing will occur later)*
* [ ] Checked for security implications? *(later)*
* [ ] Updated the documentation/wiki? *(later)*


static/js/src/entry/user-portal/App.vue
```diff
@@ -1,6 +1,10 @@
 <template>
-  <error-view v-if="context.hasError" />
-  <router-view v-else />
+  <div>
+    <nav style="height:45px;background:#fff;border-bottom:1px solid #dcdcdc" />
+    <error-view v-if="context.hasError" />
+    <router-view v-else />
+    <footer style="height:300px;background:#222;" />
+  </div>
 </template>
```
ashley-hebler: Could we use our `has-bg-white` and `has-bg-black-off` here to keep the hex values dynamic? The other inlines make sense.

AndrewGibson27: Oh lol those are completely temporary cuz I'm gonna handle the nav and footer in a different branch. Should've noted.

static/sass/ds/6-components/_appeal.scss
```diff
@@ -0,0 +1,21 @@
...
+    padding-top: $size-xxl;
+  }
+
+  &__item {
+    align-items: center;
+    display: flex;
+
+    > .c-icon {
+      margin-right: $size-xs;
+    }
```
ashley-hebler: To avoid that extra level of specificity, it seems like you wouldn't need the `>` direct child selector. Unless we foresee more deeply nested icons.

To take that a step further, you could even say `c-appeal__icon` would almost be the real way to target that and keep it low specificity, but I'm not sure if the icon component make that harder to add.

static/sass/ds/6-components/_messages.scss
```diff
@@ -0,0 +1,9 @@
+.c-messages {
+  > .c-message {
+    margin-bottom: $size-xxl;
+
+    &:last-child {
+      margin-bottom: 0;
+    }
```
ashley-hebler: Same specificity note here. I don't really see a reason for the `>`, and then having to override it again by nesting :last-child within that
```scss
.c-messages {
 .c-message {
   &:last-child {
   }
 }
}
```
seems like it'd still accomplish the same thing

static/sass/ds/6-components/_side-nav.scss
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  &__inner {
+    text-indent: 2.5rem;
+
+    > li {
+      margin-bottom: $size-b;
+    }
+  }
+}
```
ashley-hebler: Could the `has-text-{color}` helpers do some work here?

static/sass/ds/6-components/_summary-box.scss
```diff
@@ -0,0 +1,24 @@
+.c-summary-box {
+  border-bottom: 2px solid $color-white-off;
+  padding-bottom: $size-xl;
+
+  @include mq($until: bp-l) {
+    background: $color-white-pure;
```
ashley-hebler: You might get away with letting the `has-bg-white` class sit on the `c-summary-box` div and then you won't need to say this needs to be the color in your media query

static/sass/ds/7-utilities/_bg.scss
```diff
@@ -0,0 +1,19 @@
...
+
+.has-white-off-bg {
+  background-color: $color-white-off;
+}
+
+.has-white-off-bg-until-bp-l {
+  @include mq($until: bp-l) {
+    background-color: $color-white-off;
+  }
+}
```
ashley-hebler: The media query change totally make sense to me because we don't have classes for those. I'm confused about the need for renaming the `has-white-bg` and `has-white-off-bg` though. Want to chat through that?

static/js/src/entry/user-portal/routes/home-exact/Index.vue
```diff
@@ -1,44 +1,194 @@
...
+              </span>
+            </template>
+          </info-list>
+        </template>
+
+        <template v-slot:links>
+          <ul v-if="!pwResetSuccess && !pwResetFailure" class="c-link-list">
+            <li>
+              <span class="c-link-list__arrow has-text-teal">
+                <strong>&rarr;</strong>
```
ashley-hebler: In the mockup it kinda looks like this is an actual icon as opposed to an HTML entity. Not a huge deal either way, just an aesthetic thing. Maybe there's a benefit to _not_ using an SVG?

static/sass/ds/6-components/_message.scss
```diff
@@ -0,0 +1,33 @@
...
+    align-items: center;
+    display: flex;
+
+    .c-icon {
+      margin-right: $size-b;
+    }
+  }
+
+  &__close {
+    border: 1px solid $color-white-off;
```
ashley-hebler: Probably deliberate, but just in case I'll note it here: I'm not seeing the border in the mockup, but maybe it came in a later revision

ashley-hebler:
I think it's super smart to create that new folder structure. It further inspires me to start aligning the other pages with the DS. Thanks for your patience in that!

Looks wonderful! I left some design-related notes below:
I'm not seeing that teal button yet (but maybe it's just not for this round)
Totally a future UI iteration, but it'd be cool if when you clicked the sub items like Donation History and Payment History, the little yellow left-border would persist.

I think we could stand to have a `.has-list-reset {list-style: none}` helper, but that's on me. It's not hurting anything to not have it for now.

Other than that I left some nit-picky comments, which I'm happy to discuss out loud if it's easier. It's very cool to see the DS in practice and super helpful to see how you thought about it. Thanks @AndrewGibson27!

static/sass/ds/6-components/_appeal.scss
AndrewGibson27: I removed the direct-child selector. I typically add it by default because I've been burned way too many times in the past by accidentally styling deep descendants ... but I also totally get why it's probably not best practice to include it unless needed.

Re: your second point ... I'm torn. I think your way is probably the "official" BEM way to do it, but I also don't like how that way make it seem like the `.c-appeal__icon` element is a blank slate when in fact it has stying baked in via `.c-icon`. I think I prefer being explicit and letting my CSS say, "Be careful styling that nested icon!"

static/sass/ds/6-components/_messages.scss
AndrewGibson27: Done.

static/sass/ds/6-components/_appeal.scss
ashley-hebler: Makes sense to me! Child combinators as it turns out, are the same level of specificity, so not a deal breaker to use that sometimes. I know what you mean about getting burned by inheritance, but yeah I like only using those when the designation is absolutely needed.

static/sass/ds/6-components/_side-nav.scss
AndrewGibson27: Yes! Done.

static/js/src/entry/user-portal/routes/home-exact/Index.vue
AndrewGibson27: Shhhh ... I'm hoping to sneak in the HTML entity because I think it's better for accessibility + just easier overall.

static/sass/ds/6-components/_message.scss
AndrewGibson27: Yeah, I thought it looked weird without the border.

static/js/src/entry/user-portal/routes/home-exact/Index.vue
ashley-hebler: LOL, well I'd argue that being read "right arrow" probably doesn't do much for a nonsighted or vision impaired user. That's how I look at icons in general. They're usually just for visual presentation and deserve a text descriptor different from the icon itself. For instance no one wants to hear "three stacked bars" when using voiceover, JAWs etc; they want to hear "menu"

AndrewGibson27:
@ashley-hebler:
+ Per Emily Y., the teal button is likely going away
+ I agree re: keeping the yellow border on sub items, but I just followed the mockups ¯\_(ツ)_/¯"
+ Totally in favor of some sort of list reset. I keep going back to thinking we should have `list-style: none;` on the root `ul` selector. Sure, we'd have to override it for in-story bullets, but it feels like that's the exception, not the rule.

One small quirk I'm realizing about our BEM/helper class hybrid system (not _at all_ doubting it -- I love it overall) is it can be difficult to realize what helper classes are actually part of a component. 

Take this manufactured example:
`<section class="c-summary-box has-bg-white has-xl-btm-marg">`

In the above, `has-bg-white` is part of that component. All summary boxes should white backgrounds. The margin helper, though, could be adjusted or dropped entirely depending on where a box lives. Yet you wouldn't know this difference by just looking at that element.

I think maybe the solution is just to ensure that in the DS HTML examples, we only include helpers that should always go side by side with a particular `c-` class.

ashley-hebler:
> In the above, `has-bg-white` is part of that component. All summary boxes should white backgrounds. The margin helper, though, could be adjusted or dropped entirely depending on where a box lives. Yet you wouldn't know this difference by just looking at that element.

Hmm I see what you mean, BUT _should_ all summary boxes have white backgrounds? If a summary box appeared on a white background, you wouldn't go out of your way to set it. There are times when we'd want to bake in a bg color though; I'm not opposed to that.

AndrewGibson27:
> > In the above, `has-bg-white` is part of that component. All summary boxes should white backgrounds. The margin helper, though, could be adjusted or dropped entirely depending on where a box lives. Yet you wouldn't know this difference by just looking at that element.
> 
> Hmm I see what you mean, BUT _should_ all summary boxes have white backgrounds? If a summary box appeared on a white background, you wouldn't go out of your way to set it. There are times when we'd want to bake in a bg color though; I'm not opposed to that.

I mean maybe that wasn't the best example, but I think the helper-class-that-always-belongs-on-component situation is still one that'll continue coming up.

PR #206, opened 2019-05-24
Remove font-awesome and replace with svgs

ashley-hebler:
#### What's this PR do?

Swaps font-awesome spans with our `c-icon` svg markup

Icon fonts have some perf implications and we're moving to SVGs in our other repos.

`make`
`yarn run dev`
`make restart` for good measure

- Check the footer social links
- Check the /circle page stars
- Make sure nothing barks at you in the build

N/A

I'm using a macro. I don't see those used anywhere else so if we're not into macros, I'm happy to change. It felt like a lot to use [jinja's `with` syntax](http://jinja.pocoo.org/docs/2.9/templates/#with-statement), so a macro _looked_ more concise. I'm just not sure if there's some negative impact of using macros and that's why we haven't implemented.

https://3.basecamp.com/3098728/buckets/736178/todos/1781295778

#### TODOs / next steps:

* [ ] *Remove _fontawesome.scss file (keeping it there for now just in case we find we need it in the short term)*


.gitignore
```diff
@@ -19,3 +19,4 @@ yarn-error.log
 # SVG sprite
 /templates/includes/base_icons.html
 **/__pycache__
+.DS_Store
```
AndrewGibson27: Minor, but might be good to use of [these two](https://github.com/texastribune/texastribune/blob/need-for-speed/.gitignore#L2-L3) strategies for those pesky DS_Store files.

config/paths.js
```diff
@@ -4,6 +4,7 @@ const glob = require('fast-glob');
 const { replaceExtension } = require('ds-toolbox-test/tasks/utils');
 
 const CSS_OUTPUT_DIR = './static/build/';
+const SVG_LIB_DIR = './node_modules/ds-toolbox-test/assets/icons/base/';
```
AndrewGibson27: No need to update here, we might be better off constructing URLs with `path.join()/resolve()` going forward. I think they're more bulletproof in case files move.

ashley-hebler:
> LGTM! As we've already talked about, we'll want a larger strategy for bringing the DS into this app. But I think bringing in just the pieces needed for icons right now is just fine.

Yeah didn't want this one to snowball and I'm waiting on those layout class updates which is waiting on Need for Speed which is waiting on User Profile 😆

PR #208, opened 2019-05-28
Add optional inputmode attr

ashley-hebler:
#### What's this PR do?

Adds an `inputmode` prop to our <TextInput/> component to render that attribute on text fields where the prop is passed.

`inputmode` is a tiny UX enhancement for supported touch devices. Designating the `inputmode` sets the appropriate touch keyboard for a user and saves them from having to toggle to the correct character set when filling out our forms.

Inspired by this [CSS Tricks blog post](https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/)

> Unlike changing the type of the form, inputmode doesn’t change the way the browser interprets the input — it instructs the browser which keyboard to display.

Spin up locally and ensure the following:

- All Zip code fields: `inputmode="numeric"`

- All email fields: `inputmode="email"` (I think `type="email"` already provides the correct keyboard, so this one isn't completely necessary, but I figure can't hurt)

- All url fields: `inputmode=url`

- Donation amount field on /donate: `inputmode="decimal"` I didn't use numeric here because I'm assuming our form validation doesn't accept anything but numbers and decimals. This forces a number + decimal pad disallowing changing to letters, punctuation, etc. I feel fine about this since it matches the validation rules, but open to discuss.

- All remaining fields don't have the `inputmode` attribute at all.  (This is why you see `default=null` in the prop type definition. If there's a better way to force the attr to not show, let me know.)

Try it out on device and click around the form with this link: https://donations-testing-pr-208.herokuapp.com/donate

N/A

Shouldn't be. 

https://3.basecamp.com/3098728/buckets/736178/todos/1803780167

### DIFF
--- a/static/js/src/elements/TextInput.vue
+++ b/static/js/src/elements/TextInput.vue
@@ -10,6 +10,7 @@
       :name="name"
       :placeholder="placeholder"
       :type="type"
+      :inputmode="inputmode"
       @input="updateSingleValue($event.target.value)"
     />
     <p v-if="showError && !isValid" role="alert">{{ message }}</p>
@@ -49,6 +50,12 @@ export default {
     type: {
       type: String,
       default: 'text',
+    },
+
+    inputmode: {
+      type: String,
+      default: null,
+      required: false,
     },
   },
 
--- a/static/js/src/entry/business/TopForm.vue
+++ b/static/js/src/entry/business/TopForm.vue
@@ -38,6 +38,7 @@
             type="email"
             base-classes="form__text form__text--standard"
             name="stripeEmail"
+            inputmode="email"
           />
         </div>
       </div>
@@ -59,6 +60,7 @@
             label-text="website"
             base-classes="form__text form__text--standard"
             name="website"
+            inputmode="url"
           />
         </div>
       </div>
@@ -101,6 +103,7 @@
             label-text="zip code"
             base-classes="form__text form__text--standard"
             name="shipping_postalcode"
+            inputmode="numeric"
           />
         </div>
       </div>
--- a/static/js/src/entry/circle/TopForm.vue
+++ b/static/js/src/entry/circle/TopForm.vue
@@ -37,6 +37,7 @@
             type="email"
             base-classes="form__text form__text--standard"
             name="stripeEmail"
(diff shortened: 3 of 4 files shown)


static/js/src/elements/TextInput.vue
```diff
@@ -50,6 +51,12 @@ export default {
       type: String,
       default: 'text',
     },
+
+    inputmode: {
+      type: String,
+      default: null,
```
AndrewGibson27: I typically use `false` when I want an attribute not to appear. But it doesn't matter much.

ashley-hebler: I saw that vue docs said to be empty it could be `null`, `undefined`, `false`, or `disabled` so I had a 25% of getting it right. I'll change mine to keep it consistent

ashley-hebler:
> FWIW, looks like `ds-toolbox-test` is in `yarn.lock` twice on this branch. I noticed because when I did a fresh `yarn`, it tried to remove the one from `#master`.


Is there a chance that's some quirk that might be happening in your local env? I'd think if anything were specific to this branch, I'd see a code change in `yarn.lock` and it listed as a file here:
Screen Shot 2019-05-31 at 3 45 39 PM

But maybe I somehow snuck it in another way that I wouldn't see in git? I'm confused :sadtodd:

AndrewGibson27:
@ashley-hebler Yeah, maybe it has been this way for a little while. [Here's where](https://github.com/texastribune/donations/blob/inputmode/yarn.lock#L2812-L2846) I see it.

ashley-hebler:
I see I see, [_you_ removed the dupe here](https://github.com/texastribune/donations/commit/a564ead67a7c40a46adbd9dfa89d1117578f8f09#diff-8ee2343978836a779dc9f8d6b794c3b2L2812) and I haven't rebased since then. I'll force push just to be safe, but I don't think that would have added an extra instance.

static/js/src/elements/TextInput.vue
ashley-hebler: Going to leave this as null for now. ESlint doesn't like false because it's a boolean and I hate to add another type just for that.

PR #212, opened 2019-05-31
Build portal navbar using design system

AndrewGibson27:
#### What's this PR do?
+ Creates a navigation Vue component
+ Styles it with the DS `c-navbar` partial (see [related PR](https://github.com/texastribune/ds-toolbox/pull/26))

An app is not complete without navigation!

+ `make`
+ `make restart`
+ Go to `/user-portal`
+ Confirm there's a white-background nav that has a functioning dropdown on small screens

NA.

Less! Cleaning up the nav CSS felt good.

UMP.

* [ ] Added automated tests?
* [x] Tested manually on mobile?
* [x] Updated the documentation/wiki? *(added comments in DS repo)*


static/js/src/entry/user-portal/components/Icon.vue
```diff
@@ -1,6 +1,12 @@
 <template>
-  <span class="c-icon" :class="`t-size-${display.size}`">
-    <svg aria-hidden="true"><use :xlink:href="`#${name}`" /></svg>
+  <span
+    class="c-icon"
+    :class="`t-size-${mergedDisplay.size} has-text-${mergedDisplay.color}`"
```
ashley-hebler: I always like the escape hatch of being able to sprinkle in those extra classes, but I'm all for keeping it simple until we need it.

ashley-hebler:
What a world of a difference a nav makes. This is looking awesome! All worked great for me @AndrewGibson27

PR #213, opened 2019-06-03
UMP [DO NOT MERGE -- HEROKU DEPLOY NEEDED ONLY]

AndrewGibson27:


emilyyount:
@AndrewGibson27: I pushed tweaks to sizing/spacing. Also removed member benefits from the blast pages. When do you think Ben's icon svgs will be added? I imagine I'll need to tweak padding a bit once the real ones are there.

PR #224, opened 2019-06-05
Update footer to latest ds-toolbox utility classes

ashley-hebler:
#### What's this PR do?
Syncs up some updates @AndrewGibson27 and I have been making in our CSS framework for UMP and otherwise with the current site footer.

This is just another iteration of our overall CSS refactor efforts

```
make backing
make
yarn run dev
make restart
```
Make sure the footer doesn't look bonkers

N/A

Adds 3kb of CSS, so not the end of the world, but once we utilize more of the helper classes in other places, we can probably cut down the repo specific CSS to counter this increase.

Nothing specific, but improvements came as a byproduct of the UMP portal work

#### Have you done the following, if applicable

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/static/sass/6-components/_footer.scss
+++ b/static/sass/6-components/_footer.scss
@@ -1,14 +1,5 @@
 .c-site-footer {
-  
-  &__logo {
-    font-size: 4rem;
-    margin-bottom: 2.6rem;
-  }
   &__icon {
     margin-right: $size-xxxs;
-  }
-  // To be replaced with DS global styles
-  &__list {
-    list-style: none;
   }
 }
--- a/static/sass/6-components/_footer.scss
+++ b/static/sass/6-components/_footer.scss
@@ -1,6 +1,14 @@
 .c-site-footer {
-  .border--yellow_notch {
-    border-top: 2px solid $color-yellow-tribune;
-    margin: $size-xs 0;
+  
+  &__logo {
+    font-size: 4rem;
+    margin-bottom: 2.6rem;
+  }
+  &__icon {
+    margin-right: $size-xxxs;
+  }
+  // To be replaced with DS global styles
+  &__list {
+    list-style: none;
   }
 }
--- a/static/sass/6-components/_footer.scss
+++ b/static/sass/6-components/_footer.scss
@@ -1,5 +1,6 @@
+// To be removed after this repo is set up to take the base ds-toolbox elements file
 .c-site-footer {
-  &__icon {
-    margin-right: $size-xxxs;
+  &__links {
+    list-style: none;
   }
 }
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "axios": "^0.19.0",
     "babel-loader": "^8.0.0",
     "css-loader": "^2.1.1",
-    "ds-toolbox-test": "https://github.com/texastribune/ds-toolbox.git#development",
+    "ds-toolbox-test": "https://github.com/texastribune/ds-toolbox.git#master",
     "es6-promise": "^4.2.4",
     "fast-async": "7",
     "fast-glob": "^2.2.6",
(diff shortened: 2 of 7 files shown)


static/sass/6-components/_footer.scss
```diff
@@ -1,6 +1,14 @@
 .c-site-footer {
-  .border--yellow_notch {
-    border-top: 2px solid $color-yellow-tribune;
-    margin: $size-xs 0;
+  
+  &__logo {
+    font-size: 4rem;
```
AndrewGibson27: Do you hate the idea of using an inline style to dictate the size of comically large icons like this one? I left a [comment](https://github.com/texastribune/ds-toolbox/pull/25#issuecomment-497519062) about that strategy in this PR -- not sure if it's a good one. Also doesn't matter much.

static/sass/6-components/_footer.scss
```diff
@@ -1,6 +1,14 @@
...
+    margin-bottom: 2.6rem;
```
AndrewGibson27: Could use a utility class here, I think. I use `has-b-btm-marg` in the portal.

static/sass/6-components/_footer.scss
```diff
@@ -1,6 +1,14 @@
...
+  }
+  &__icon {
```
AndrewGibson27: I'm torn here. Part of me thinks social icons in the footer will be common enough that the style for it should live at the DS level. Part of me thinks it'll make more sense at the repo level given, for example, the UMP footer doesn't have social icons. IDK.

AndrewGibson27:
Looks good! I noticed some two-line text gets a little jammed on small screens. That's not something I encountered in UMP because none of the anchor text is long enough. Could either drop some line-height on that `a` rule at the DS level or get crazy and stack those columns like I do on UMP. Or nothing! Or something else!

static/sass/6-components/_footer.scss
ashley-hebler: I'm torn too! That's a tough one. I like the idea of it being in the DS because that'd be the last remaining style, allowing it to only pull from the DS with no repo-specific code. Same case with the /texastribune footer. I don't like UMP getting unused CSS, but it _is_ only 1 declaration 🤔

static/sass/6-components/_footer.scss
ashley-hebler: I don't love it, but I understand what you mean. Tis changed.

static/sass/6-components/_footer.scss
ashley-hebler: good catch!

ashley-hebler:
@AndrewGibson27, thanks for taking a look. Yes line-height was not the way to go on the footer links spacing. I added something to remedy that greater issue [here](https://github.com/texastribune/ds-toolbox/pull/29). Hopefully it doesn't gross you out too much.

I can also sneak that `c-site-footer__icon` right spacing into the DS too if you like. See my reply on that and let me know what ya think. Other than those I implemented the other fixes 👍

static/sass/6-components/_footer.scss
AndrewGibson27: Up to you, but my final vote is for adding to DS. It wouldn't shock me at all if someone asks me to add the social icons to the UMP footer, in which case there will be no unused DS code!

AndrewGibson27:
This footer is of course still using the old underscore-ridden grid system. I went ahead and torched that in my UMP footer in favor of CSS grid, but I totally understand if that's out of scope for this PR.

ashley-hebler:
> This footer is of course still using the old underscore-ridden grid system. I went ahead and torched that in my UMP footer in favor of CSS grid, but I totally understand if that's out of scope for this PR.

Yes, in my mental map (that no one can see 😆) we tackle all the `grid_row`, `col` instances at the same time. And tbh, I've been putting it off because it sounds hard.

I'll add that icon rule to the DS. And then we should be all set for this. Thanks again!

PR #278, opened 2019-07-12
Add latest queso versions and update task names

ashley-hebler:
#### What's this PR do?
1. `queso-tools` and `queso-ui` move to the latest release 
2. Adds more consistent npm script command names

1. Doesn't really help us, but I like the idea of `queso-ui` staying pretty up to date. As we add more features, it'll be nice to reach for them and have them there without having to update. The build tasks got a security patch which didn't do much, but I figured update while I'm in there.

2. I hated the names I created because I couldn't remember them and they contradicted the naming convention on the js tasks. The prefix of `css:` is little misleading since it does icons too, but eh feels close enough.

`make backing`
`make`
`make restart`

Just make sure it all builds and the styles aren't broken on /donate

Shouldn't

nope

off-sprint; general housekeeping

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,13 +2,13 @@
   "name": "donations",
   "description": "Collect donations for nonprofits",
   "scripts": {
+    "css:dev": "NODE_ENV=development node config/tasks/dev.js",
+    "css:prod": "NODE_ENV=production node config/tasks/build.js",
     "js:dev": "NODE_ENV=development webpack --config webpack/config.dev.js",
     "js:prod": "NODE_ENV=production webpack --config webpack/config.prod.js",
-    "dev": "npm-run-all --parallel ds-tasks-watch js:dev",
-    "postinstall": "yarn run ds-tasks-build && yarn run js:prod",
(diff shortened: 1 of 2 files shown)


ashley-hebler:
> ...extraneous dependencies hanging around in `package.json` (or maybe I'm just missing something): glob-watcher, fast-glob, etc.

Yeah fast-glob is how we used to grab `*/**.scss` files. Now we're more explicit so that can go. I guess I can lump that in. This wasn't super tied to that, but I see why you made the connection. Glob watch is still needed.
https://github.com/texastribune/donations/blob/master/config/tasks/dev.js#L1

ashley-hebler:
@AndrewGibson27 I 🗑 ed **fast-glob** and **node-sass**, which created lots of red in yarn.lock. I mean that's good and I'd imagine expected. Just to be safe, I'll stick with deploy as infrequently as possible on Fridays rule and push this out Monday.

PR #307, opened 2019-08-15
Portal read + write

AndrewGibson27:
@ashley-hebler Obviously still a work in progress, but feel free to poke around at your convenience to see if you hate any of my code.


static/sass/ds/7-utilities/_bg.scss
```diff
@@ -9,3 +9,17 @@
...
+  }
+}
+
+.has-bg-twitter {
+  background-color: $color-twitter;
+}
+
+.has-bg-facebook {
+  background-color: $color-facebook;
+}
```
ashley-hebler: The good news is this is included in queso, so no need to add. The bad news is I failed to document it. I'll add it to the SCSS comments so it appears in the docs. I don't blame you for not seeing that one!

static/sass/ds/6-components/_message.scss
```diff
@@ -22,7 +23,7 @@
     border: 1px solid $color-white-off;
     border-radius: 50%;
     cursor: pointer;
-    padding: .25rem .45rem;
+    padding: .5rem .5rem;
```
ashley-hebler: For some reason this padding setting is looking a little oblong on my screen. I always set a width and height on circle buttons, but maybe that's old school 👵. If you go that route, `l-align-center-children` is helpful to center the little x icon.

AndrewGibson27: @ashley-hebler Yeah, I jacked that up. Should be fixed now.

ashley-hebler:
This is so much work and it's looking super slick. I didn't test functionality a whole lot, but that saved state is a nice solve. I like that the loader pauses you for a bit so you interpret the change.

Anyway I left a few notes, but I didn't speak to any of the Vue commits since you are a wizard in that, while I'm a mere muggle ⚡️

static/sass/ds/7-utilities/_bg.scss
AndrewGibson27: Removed!

PR #368, opened 2019-11-08
Allow leading dollar signs in amounts

AndrewGibson27:
#### What's this PR do?
Allows `$88` and `88` to both be valid amounts on the donate form.

Sarah G. request.

+ `make`
+ `make restart`
+ Confirm you can successfully donate with and without a leading dollar sign on `/donate`
+ To be sure, confirm you can successfully donate on `/circle` and `/business` too

NA.

I originally wanted to handle this all client side. But it was a strange and confusing UX to remove characters after someone finished typing an amount. I then thought about maintaining some sort of background data store with data that's formatted differently than what you actually see on the form (in other words, with dollar signs stripped out). But that proved nearly impossible given we literally submit the form and the data it contains (as opposed to use it as a vehicle to construct an API payload like in UMP).

Thus, I simply updated the client-side validator to allow leading dollar signs, then updated the back-end logic to remove the dollar sign (if it exists), validate the amount and then convert it to an integer.

@x110dc If you could please confirm I didn't butcher anything too much on the Python side, I'd be grateful.

https://3.basecamp.com/3098728/buckets/736178/todos/2178926818

* [x] Added automated tests?

### DIFF
--- a/forms.py
+++ b/forms.py
@@ -10,7 +10,27 @@ from wtforms.fields import (
 from wtforms.fields.html5 import EmailField
 
 
-def strip_filter(value):
+def validate_amount(form, field):
+    value = field.data
+    if value is None:
+        raise validators.ValidationError("Invalid amount")
+    if value < 1:
+        raise validators.ValidationError("Amount is less than 1")
+
+
+def format_amount(value):
+    if value.startswith("$"):
+        value_to_convert = value[1:]
+    else:
+        value_to_convert = value
+    try:
+        float(value_to_convert)
+        return float(value_to_convert)
+    except ValueError:
+        return None
+
+
+def strip_whitespace(value):
     if value is not None and hasattr(value, "strip"):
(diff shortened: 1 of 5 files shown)


forms.py
```diff
@@ -10,7 +10,30 @@
...
+# if value starts with a dollar sign, remove it
+# then convert to a float
+def format_amount(value):
+    if value.startswith("$"):
+        value_to_convert = value[1:]
+    else:
+        value_to_convert = value
+    try:
+        float(value_to_convert)
+        return float(value_to_convert)
```
x110dc: Not a big deal but you could replace these with:
```
       value = value[1:]
try:
   return float(value)
```
The return is in the `try` block so you don't need to do the cast twice.

AndrewGibson27:
@x110dc I took my best crack at unit tests. Let me know if they're utter garbage.

PR #424, opened 2020-01-31
Improve UMP data-store management

AndrewGibson27:
#### What's this PR do?
+ Makes our Vuex store much more of a "one source of truth" for the app's state, primarily by using lots o' [getters](https://vuex.vuejs.org/guide/getters.html)
+ Creates a [scoped-slot](https://vuejs.org/v2/guide/components-slots.html#Scoped-Slots) provider and [mixin](https://vuejs.org/v2/guide/mixins.html) for each piece of the store (for easy access inside components)
+ Un-nests many components so they're more reusable across routes
+ Greatly improves the view-as feature
+ Cleans up the ESLint config by now allowing dynamic import statements
+ camelCases everything

While this app has always used Vuex, v1 didn't do a great job of truly maxing out its potential. Instead, I was doing lots of data mangling and combining at the component level, which lead to some duplicate logic across routes and thus was sort of defeating the purpose of a one-source-of-truth data store. For example, it's likely we will at some point want to show someone's transaction history in multiple places throughout the portal. Right now, building that list is done at the [component level](https://github.com/texastribune/donations/blob/master/static/js/src/entry/account/routes/payments/containers/DetailContainer.vue#L23-L53). After this PR, it'll be done at the [store level](https://github.com/texastribune/donations/blob/ump-store/static/js/src/entry/account/store/user/index.js#L253-L261), meaning using it across routes is just a matter of dropping in a mixin and calling it a day.

The view-as feature has also been pretty brittle for a while now. There are some routes in which it throws an error, and the person you're viewing-as must have an Auth0 account. Now, it works on all routes, and there's no account requirement. Sarah Glenn will be happy!

Go to `/account/`, log in, click around all the routes. Ensure nothing looks amiss.

As with all UMP changes, I'll spend time viewing-as a bunch of different donors post-deploy to ensure everything looks OK. I'm quickly learning that's the quickest way to test portal changes.

No user-facing changes.

Because this changes the Vuex store -- aka where all user data is held -- it required updating all routes and many components. So there are a lot of changed files. I'll ensure I do my due diligence in testing this.

https://3.basecamp.com/3098728/buckets/736178/todos/2352590914

* [ ] Added automated tests?
* [ ] Checked BrowserStack?
* [ ] Checked accessibility?


static/js/src/entry/account/components/InfoList.vue
```diff
@@ -1,13 +1,13 @@
...
       <h3 class="t-size-s has-xxxs-btm-marg">
         <strong>{{ item.heading }}</strong>
       </h3>
       <p>
-        <slot :item="item">
-          <span class="has-text-gray-dark">{{ item.text }}</span>
-        </slot>
+        <span class="has-text-gray-dark">
+          <slot name="text" :item="item"> {{ item.text }} </slot>
+        </span>
```
ashley-hebler: Is the extra span needed? Could it be just a p with a slot?
```
<p class="has-text-gray-dark>
    <slot...
</p>
```

AndrewGibson27: Yes that is better!

PR #431, opened 2020-02-14
Improved UMP data fetching and error handling

AndrewGibson27:
#### What's this PR do?
+ Moves data fetching from the component level to the router level
+ Allows routes to wait for their parent route to finish fetching before they do their own fetching
+ Moves redirects outside components and into `beforeEnter` router hooks
+ Creates new error classes that can accept arbitrary metadata and a status code
+ Configures Sentry reporting to accept that additional data
+ Simplifies our error view
+ Simplifies page-view logging

Right now, we have a messy in-component mixin that allows child routes to wait for their parent route before doing data fetching. This is messy and complicated as it requires passing down a `parentRouteIsFetching` prop and showing lots of loading wheels along the way. By moving data fetching out of components entirely, navigation won't be confirmed until data has been fetched. In other words, when a new route loads, it'll already have all the data it needs: no need for additional loading indicators. Additionally, it means the user won't see the "flash" of a route they're not supposed to be on (for example, if someone tries to visit `/account/blast/` when they're not a Blast subscriber).

Improving error handling is beneficial in lots of ways. Obviously, it gives us more detail about problems users might be encountering. But by associating each error type with a status code, it means that if/when we want to do server-side rendering, we can send back proper status codes instead of 200s every time.

First, go to 1pass and find the note called "Sentry PR review" (I'm not sure what Sentry values are secret, so I just put them all in here). Copy those into your `env-docker`, replacing <your-name> with, you guessed it, your name.

Then, go to the `donations` project on our `sentry.io` site and filter by the environment specified above. Keep that tab open. Next:
- `make`
- `make restart`
- Go to `/account/` and click around a few routes, just looking for anything that's obviously broken
- Assuming you haven't signed up for The Blast with your test Auth0 email address, put `/account/blast/` in your address bar. Confirm that you eventually get redirected back to the account overview.
- Without refreshing, open up your browser console and run `console.log(window.dataLayer);`. There should only be one `userPortalPageview` event, and it should be for "Account Overview," not "The Blast."
- Open up `static/js/src/entry/account/index.js` and change `!isVerified` to `isVerified` on line 197 (we're going to force an error to occur). Wait for Webpack to rebuild the bundles, then hard refresh. You should get sent to an error screen. Wait a few seconds, then hard refresh that Sentry tab you've got open. There should be a new error logged in there -- something about accessing a protected route while not verified. If you click into that error and scroll down to `Additional Data --> status`, it should be a 403.
- Undo the above change and wait for Webpack to rebuild
- Lastly, log out via the top-right nav button, then put `/account/` back in your browser. You should get redirected to a log-in screen.

NA.

LESS.

https://3.basecamp.com/3098728/buckets/736178/todos/2352589312
https://3.basecamp.com/3098728/buckets/736178/todos/2352588307

* [ ] Added automated tests?
* [ ] Checked BrowserStack?
* [ ] Checked accessibility? 


ashley-hebler:
All those steps WFM!

PR #434, opened 2020-02-21
Improve UMP token handling

AndrewGibson27:
- Refactors our Vuex store module for handling the Auth0 user
- Distinguishes between being logged out of Auth0 vs. being logged in but having an Auth0 error
- Makes it so we're only fetching new tokens if one is about to expire (as opposed to every 15 minutes)
- Adds a conditional view to the "Confirm linked identity" route to account for someone who is logged into a different account than the one from which they initiated the email link (I came to realize this was a problem from this [Sentry ticket](https://sentry.io/organizations/the-texas-tribune/issues/1524248064/?environment=production&project=1304832&query=is%3Aresolved&statsPeriod=14d))
- Refactors the user portion of the Vuex module to account for some portal API changes per [this PR](https://github.com/texastribune/portal-api/pull/162)


ashley-hebler:
I kinda just clicked around the portal and that all WFM. I did [hit a error](https://sentry.io/organizations/the-texas-tribune/issues/1526201777/?environment=Ashley&project=1304832&query=is%3Aunresolved) when I tried to update info, but it kinda seems like that was just a weird network glitch. I worked fine after refreshing.

PR #454, opened 2020-03-31
Robots.txt Add disallow for another account endpoint

valasensio:
#### What's this PR do?
Disallow directive added for `/account`

Bots hitting `/account` which results in an error

Check if the `Disallow: /account`  line is included in robots.txt 
We'll see if the error goes away in Google Search Console &/or Alexa, but we can't truly control bots, just ask them nicely to stop

Not needed

Nah

[Add /account to Donations robots.txt](https://3.basecamp.com/3098728/buckets/736178/todos/2491237469#__recording_2542145848)

N/A

#### TODOs / next steps:

* [ ] I'll follow up and see if we're no longer  seeing robots hitting `/account`

### DIFF
--- a/robots.txt
+++ b/robots.txt
@@ -1,2 +1,3 @@
 User-agent: *
 Disallow: /account/*
+Disallow: /account


valasensio:
@AndrewGibson27 Not a high-priority, just whenever

PR #459, opened 2020-04-03
Run npm install to sync lock file with latest

ashley-hebler:
#### What's this PR do?

This is the output of running `npm install` on master

The hope is that every time you need to run npm install, you're not committing an unintentional change.

On master:
run `make` to get into the app and then `npm install`. `git status` should show a change in package.lock.json. Undo those changes to clean your staged commits.

In this branch:
Run `npm install` again. Confirm that package.lock.json remains unchanged.

N/A

This may introduce a security warning about acorn in the GitHub interface. If that's the case we'll push up another PR with a manual `npm audit fix` and deal with that in each subsequent PR.

N/A - Just trying to prevent weird future commits

#### TODOs / next steps:

* [ ] *Determine if this creates a new security warning and run `npm audit fix` if so*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3355,7 +3355,8 @@
       "dependencies": {
         "acorn": {
           "version": "6.4.0",
-          "resolved": "",
+          "resolved": "https://registry.npmjs.org/acorn/-/acorn-6.4.0.tgz",
+          "integrity": "sha512-gac8OEcQ2Li1dxIEWGZzsp2BitJxwkwcOm0zHAJLcPJaVvm58FRnk6RkuLRpU1EujipU2ZFODv2P9DLMfnV8mw==",
           "dev": true
         },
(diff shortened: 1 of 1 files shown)


AndrewGibson27:
:shipit:

ashley-hebler:
@AndrewGibson27, I've been trying to verify that donations still submit properly on this. I'm getting successful submit, but it never seems to hit the test Slack channel. Should I be worried about that? I'm almost sure it's some env var I'm missing - looks like September was the last time that ever worked for me. I have `ENABLE_SLACK`, `SLACK_API_KEY`, and `SLACK_CHANNEL` all filled out so I'm not sure what's missing there.

ashley-hebler:
Adding some Slack context to my previous comment. We determined this to be a bug and we'll consult @x110dc on what could be happening there.

PR #467, opened 2020-04-17
add clack notifications on failed charge

tjb346:
#### What's this PR do?
Adds slack notifications on charge errors and removes success notification in #donations on error.

For charges where there were no notification before I did not add new ones because I don't know if that was done for privacy reasons, etc.

Better understanding via slack what happened with donations.

Check slack for notification of an invalid charge.

N/A

No

Off-sprint

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/app.py
+++ b/app.py
@@ -52,10 +52,10 @@ from util import (
     clean,
     notify_slack,
     send_email_new_business_membership,
-    send_multiple_account_warning,
+    send_multiple_account_warning, send_slack_message,
 )
 from validate_email import validate_email
-from charges import charge
+from charges import charge, ChargeException
 
 ZONE = timezone(TIMEZONE)
 
@@ -317,9 +317,12 @@ def add_donation(form=None, customer=None, donation_type=None):
     if period is None:
         logging.info("----Creating one time payment...")
         opportunity = add_opportunity(contact=contact, form=form, customer=customer)
-        charge(opportunity)
-        logging.info(opportunity)
-        notify_slack(contact=contact, opportunity=opportunity)
+        try:
+            charge(opportunity)
+            logging.info(opportunity)
+            notify_slack(contact=contact, opportunity=opportunity)
+        except ChargeException as e:
+            e.send_slack_notification()
         return True
 
     elif donation_type == "circle":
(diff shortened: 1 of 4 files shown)


charges.py
```diff
@@ -79,21 +95,19 @@ def charge(opportunity):
...
         logging.error(f"Problem: {e}")
-        # TODO should we raise this?
-        return
+        raise ChargeException(opportunity, "unknown error")
 
     if card_charge.status != "succeeded":
         logging.error("Charge failed. Check Stripe logs.")
-        # TODO should we raise this?
-        return
+        raise ChargeException(opportunity, "charge failed")
```
ashley-hebler, line 110: I think these would be helpful. If we start to feel they add noise to the stripe channel we can always dial back.

app.py
```diff
@@ -1078,7 +1087,11 @@ def add_blast_subscription(form=None, customer=None):
         for opportunity in opportunities
         if opportunity.expected_giving_date == today
     ][0]
-    charge(opp)
+    try:
+        charge(opp)
+    except ChargeException:
+        # TODO should we alert slack? Did not because we had no notifications here before.
+        pass
```
ashley-hebler, line 1094: hmm I wonder why we don't send these even as a success. Maybe for privacy because it's more so paying for a product; rather than a donation? In any case I don't think it hurts to send the rejected ones like we do for the others - lets us know if something fishy is going on.

ashley-hebler:
@tjb346 just saw that you tagged @AndrewGibson27 as a reviewer, but I checked this out anyway since I was already testing the updated variables. I can confirm successful slack notifies, but I wasn't sure how to force it to error. I added my 2¢ in the inline comments too. Overall I think sending failed charges #stripe will be a huge help. Thanks for doing this.

tjb346:
@ashley-hebler I added Andrew because you had said you could test locally, and that was before Andrew confirmed he also couldn't test locally. This works.

app.py
tjb346: I was thinking it might be for privacy too. I think it's probably ok, but just in case I'm going to add a question for Daniel. It's easy enough to add later once he confirms.

PR #474, opened 2020-04-30
Update Salesforce opportunities with gross and net values

x110dc:
#### What's this PR do?

Updates Salesforce to use the gross donation amount as the primary amount in opportunities. Also captures net value.

For accounting consistency. For most (all?) other revenue types the gross value is represented in Salesforce opportunities. For Stripe donations it depended on whether the donor chose to pay fees or not whether the gross or net amount was the pprimary "Amount" on the record. Now that "Amount" will always be gross, regardless. And net will also be captured so that fees can be calculated in SF if wanted. This is a request from the Evans.

Make a single test donation and choose not to pay fees. Make another donation of the same amount and choose to pay fees. Look up the Salesforce records. Confirm that the value shown in "Amount" is field is higher for the one where fees were paid by the donor. And confirm that the values for "Net" and "Donor Selected Amount" are correct. 

Make sure that Tommie, Evan Lambert and Anna know. 

No.

No ticket.

* [x] Added automated tests? *(updated)*

#### TODOs / next steps:

* [ ] After this is merged we'll need to run a job to update all previous donations since 2020-01-01 in SF. This should probably happen shortly after this is merged so reports will be consistent.


AndrewGibson27:
@x110dc I did $35 for both test donations. The front end indicates the fee amount is $1.09; this is reflected in the SF record for the with-fees donation:
Screen Shot 2020-05-04 at 6 22 01 PM

But for the without-fees donation, it looks like the difference is only $1.07. Is this expected?
Screen Shot 2020-05-04 at 6 22 38 PM

x110dc:
Yes, that's expected. [This page](https://support.stripe.com/questions/passing-the-stripe-fee-on-to-customers) explains it way better than I ever could: "taking into consideration that increasing the final amount also increases the Stripe fee".  

So for the one where you don't agree to pay fees, the fees are calculated as $35 * 0.022 - 0.30 = $1.07. And the net is $35-$1.07 = $33.93.

But when you do agree to pay fees the calculation for the gross is: ($35.00 + $0.30) / (1-0.022) = $36.09406953 which rounds down to $36.09. So the fees are $36.09-$35 = $1.09.  

You can also try with a test Amex card to see how it differs (where we charge fees as though you're paying with a regular card) and the actual fees are 3.5% (no fixed fee).

AndrewGibson27:
Thank you for the explanation! Well, looks like the values I screen shotted are correct then. If they LGTY, this LGTM!

Let me know how I can help with next steps. I see there's some stakeholder communication to be done and a job to be written and run after merge. Happy to help with whatever.

PR #490, opened 2020-05-22
Append temp message to all stripe errors

ashley-hebler:
#### What's this PR do?

see title

just adds a string to the error message we get back from stripe

I tested cards from this section https://stripe.com/docs/testing#cards-responses

Tell engineering channel

This is just temporary until we hear back from stripe

Should we be more specific about when we append the message? I wasn't sure how to do that but open to discussing that

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/app.py
+++ b/app.py
@@ -52,7 +52,8 @@ from util import (
     clean,
     notify_slack,
     send_email_new_business_membership,
-    send_multiple_account_warning, send_slack_message,
+    send_multiple_account_warning,
+    send_slack_message,
 )
 from validate_email import validate_email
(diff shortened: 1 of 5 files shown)


app.py
```diff
@@ -363,6 +364,8 @@ def do_charge_or_show_errors(form_data, template, bundles, function, donation_ty
         body = e.json_body
         err = body.get("error", {})
         message = err.get("message", "")
+        if message:
```
tjb346, line 367: You probably want to do this if there is a message or not correct?

PR #491, opened 2020-05-28
Improve error handling and give Sentry less user information

AndrewGibson27:
#### What's this PR do?
Makes it so we give less about the request/response and a logged-in user's information to Sentry.

It's always good store as little information about users as possible, especially in third-party services.

Ensure you have this in your `env-docker`:
```
ENABLE_SENTRY=True
SENTRY_DSN=https://d776f8ab57804e8a8d0f581ed0e4d3b2@sentry.io/1304832
SENTRY_ENVIRONMENT=<your name>
```
Then:
- `make`
- `make restart`
- Go to `/account/` and enter `bleh` in the fixed-position "view as" box on the left
- You should get an error page, and if you go to Sentry under your environment, there should be a new issue. Confirm the only info included under USER is IP address and user ID.

NA.

NA.

https://3.basecamp.com/3098728/buckets/736178/todos/2638868264

* [ ] Added automated tests?
* [ ] Checked BrowserStack?
* [ ] Checked accessibility?
* [x] Checked for security implications?


ashley-hebler:
WFM! Did something change where I have "View as" now? I never had that before

PR #523, opened 2020-08-06
better stripe descriptions

x110dc:
#### What's this PR do?

- In cases where the Opportunity description is blank then generate a Stripe description based on the Opportunity Type.
- Removes leading "The " from legacy donations before sending Opportunity descriptions to Stripe.

It makes things more consistent for the data in Stripe (all sustaining memberships will have the same description). And that makes things easier for accounting.

- Either spin up the server and make a single donation or create one manually in Salesforce (former is probably easier for most).
- If you created it through the server then go into Salesforce, find the Opportunity and blank out the description field. If you're creating it manually just don't enter a description.
- `python batch.py`
- Check Stripe and confirm that the description field there reads "Texas Tribune Membership"
- Create another donation by one of the two methods above. Enter "The Whatever" for the description in Salesforce.
- `python batch.py`
- Confirm in Stripe that the donation shows a description of "Whatever"
- If you have questions about any of the above ping me or Anna.

Just let Claire and Tommie know.

I don't think so.

A wiki question.

* [x] Added automated tests? *( )*

### DIFF
--- a/charges.py
+++ b/charges.py
@@ -48,6 +48,35 @@ def quantize(amount):
     return Decimal(amount).quantize(TWOPLACES)
 
 
+def generate_stripe_description(opportunity) -> str:
+    """
+    Our current code populates the Description field of recurring donations 
+    and opportunities when those are created. Those descriptions get passed 
+    on to Stripe when the card is charged. But we have at least two cases 
+    where the Description field could be blank: when someone manually enters 
+    a donation or when it's a donation that's been migrated from our legacy 
+    (Tinypass) system. But in those cases we know the opportunity type and 
+    it's a direct relationship to the description so we can populate it anyway.
+    """
+    # remove leading "The " from descriptions for better Stripe
+    if opportunity.description:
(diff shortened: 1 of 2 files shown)


AndrewGibson27:
@x110dc I'm having trouble getting a donation to show up in Stripe as anything but `Texas Tribune Membership`. If you check the test Salesforce instance, you can see I made a $12 donation as Andrew Blubber. I changed the description to `The Whatever` there, but running `python batch.py` isn't changing the description in Stripe.

x110dc:
Hmm. I'm not sure what could've happened here. Is there any possibility that the `batch.py` ran before the change to the Salesforce entry was made? I just reset your that opportunity (by changing the stage to `Pledged` and the date to today) and re-ran and it showed in Stripe as `Whatever`. Then I duplicated that opportunity, did the same and again it showed as `Whatever`. Think you could try again when we're both online?

AndrewGibson27:
WFM

PR #548, opened 2020-09-14
Improve charge error handling

tjb346:
#### What's this PR do?
Makes sure all errors during charges close the opportunity so they don't end up on wall. Sends slack messages for failed recurring donations.

To help prevent failed donations from getting on wall and to better investigate errors going forward.

Make a donation that will fail to check that we get it in Slack.

Notify SF people that errors reason will change slightly in SF from "unknown failure" to "card declined for unknown reason" if a card is declined and Stripe doesn't say why and just to "unknown failure" for everything else.

Still don't know why recurring donations may end up on wall, but that's another problem.

Off-sprint

### DIFF
--- a/batch-tmp.py
+++ b/batch-tmp.py
@@ -102,9 +102,9 @@ def charge_cards():
         )
         try:
             charge(opportunity)
-        except ChargeException:
-            # TODO should we alert slack? Did not because we had no notifications here before
-            pass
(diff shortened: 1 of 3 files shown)


AndrewGibson27 (approved):

tjb346:
Checked with Jenni that we could update reason messaging.

PR #585, opened 2020-11-28
Bad actor

x110dc:
#### What's this PR do?

Checks for bad actors. 

To make it harder for bad actors to test stolen card numbers against our site.

- Use the review app: https://donations-testing-pr-585.herokuapp.com/ (also CNAMEd to https://bad-actor-test.texastribune.org) 
- Try one-time, recurring, Circle transactions. Try to make some of them trigger the quarantine (see wiki doc for what would do that).
- Check the #test-quarantine-donation Slack channel and approve or reject donations. Confirm that they end up that way in Salesforce.

Let Sarah and anyone who might be approving transactions know and invite them to the approver channel. 

I don't think so.

* [x] Added automated tests? *( )*
* [x] Updated the documentation/wiki? *(https://texastribune.atlassian.net/wiki/spaces/TECH/pages/907018241/Bad+Actors+card+testing?moved=true)*

#### TODOs / next steps:

Check the above wiki page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,6 @@ Environment
 | `RECAPTCHA_SITE_KEY`        |                             stringstringstring |
 | `BAD_ACTOR_API_URL`         |                            https://foo.bar.api |
 | `BAD_ACTOR_NOTIFICATION_URL`|                            https://foo.bar.api |
-| `AUTH0_AUDIENCE`            |                            https://foo.bar.api |
 | `AUTH0_PORTAL_M2M_CLIENT_ID`|                                         string |
 | `AUTH0_PORTAL_M2M_CLIENT_SECRET`|                                     string |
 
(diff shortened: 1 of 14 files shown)


AndrewGibson27:
I added the Clearbit JS.

mailbackwards:
I tried a bunch of tests on single+recurring+circle, using the test Stripe test Visa on the Heroku staging URL, and from a user perspective they always "went through" and seemed to work, piping into the quarantine. A couple of notes to make sure I'm not doing anything horribly wrong:

- I couldn't get any of them to _not_ get rejected by Clearbit, but maybe I was always doing something phony or fishy (phishy?). Or I might be missing something.
- The Salesforce links are helpful but I don't know exactly what I'm doing when reading SF, so please check my work and make sure they are getting rejected as expected-- I think saying "it's closed" is the right thing in these cases though.
- I sometimes got timeout notices from Slack about waiting for SF to reply, but it would always ultimately work and give me the nice "Donation rejected" message, so maybe that's fine.

PR #766, opened 2021-08-20
use different image for better Python+Node interaction

x110dc:
#### What's this PR do?

Changes the Dockerfile to use a different image

Better handles Python+Node together.

- `make`
- `make restart` 
- Make sure frontend things happen and work as expected. 
- `make test`
- Make sure CI works. 

N/A

It relinquishes some control over the image, but the versions are pinned. 

Request on wiki. 

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,21 @@
-FROM python:3.9
+#FROM python:3.9
+FROM nikolaik/python-nodejs:python3.9-nodejs12
 
 # This Dockerfile is intended for development. See Dockerfile.sample for something more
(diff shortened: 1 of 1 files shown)


bfreeds (approved):
Thank you @x110dc, this works for me -- I was wondering about multi-stage builds or removing `nodesource` and using `apt install nodejs npm` directly, but then we'd have to use nvm or something to use the specific node version we configure the image with.  I like this simple and straightforward approach -- good find on the new docker base image.

PR #813, opened 2021-11-03
Update config.py

matthewdylan:
* update vars to match expected behavior that will be required in celery 6 (https://docs.celeryproject.org/en/v5.1.2/userguide/configuration.html#new-lowercase-settings)
* adding "redis_backend_health_check_interval" var to hopefully fix error 110 bug introduced in redis 3.2.0

#### What's this PR do?
*updates config variables to the format listed [here](https://docs.celeryproject.org/en/v5.1.2/userguide/configuration.html#new-lowercase-settings)... yes, it's early, but I was already here doing the next thing so I figured I should tackle now when I'm aware of it.
*add "redis_backend_health_check_interval" var to hopefully fix Error 110 connection bug that keeps randomly popping up as suggested [here](https://docs.celeryproject.org/en/v5.1.2/userguide/configuration.html#redis-backend-health-check-interval)

The lower case config vars is basically future proofing since we'll have to do it either way by the time we upgrade to celery 6. Adding the health_check_interval should knock off the annoying Error 110 bug.

```
make backing
make
make restart
```
ensure you can make a donation

Doesn't need to be.

The config.py file is required for the celery setup, but we've used it as a catchall for other constants used in the donations app. I think that's fine, but it does look a little odd, now that the celery-related vars are lowercase (sidenote, why did celery change these to be lowercase???) More picky than anything, but worth noting. To re-state: the lowercase config vars aren't a required update until celery 6, but makes sense to do it while I'm here since it's functional with celery 4 (where we're currently at). If the Error 110 bug persists after this, we can remove the health check and reassess.

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/app.py
+++ b/app.py
@@ -185,9 +185,9 @@ app.secret_key = FLASK_SECRET_KEY
 
 app.config.from_pyfile("config.py")
 app.config.update(
-    CELERY_ACCEPT_CONTENT=["pickle", "json"],
-    CELERY_ALWAYS_EAGER=False,
-    CELERY_IMPORTS=("app", "npsp", "batch"),
+    accept_content=["pickle", "json"],
+    task_always_eager=False,
+    imports=("app", "npsp", "batch"),
 )
 stripe.api_key = app.config["STRIPE_KEYS"]["secret_key"]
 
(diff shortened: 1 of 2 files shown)


ashley-hebler (approved):
Wow, nice sleuthing on this one. And good call to just go ahead and get us compliant with the new version. Test steps WFM!

PR #835, opened 2021-12-13
Switch notification channel batch

matthewdylan:
#### What's this PR do?
Reverts my last pr for a simpler process.

As far as functionality, this only switches the notification to send from the #donations channel to the #stripe channel. Code-wise, I reverted my updated to notify_slack which were attempting to customize it more and more. This func, as I've since found out, really should be solely used for notifying slack wrt/ a single donation. I toyed with renaming the func to make this more clear, but any names I could think of seemed really clunky.

https://3.basecamp.com/3098728/buckets/736178/todos/4347915712

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/batch.py
+++ b/batch.py
@@ -8,7 +8,7 @@ from pytz import timezone
 from charges import ChargeException, QuarantinedException, amount_to_charge, charge
 from config import ACCOUNTING_MAIL_RECIPIENT, LOG_LEVEL, REDIS_URL, TIMEZONE
 from npsp import Opportunity
-from util import send_email
+from util import send_email, send_slack_message
 
 zone = timezone(TIMEZONE)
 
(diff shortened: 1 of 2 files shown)


ashley-hebler (approved):
This LGTM. Ultimately I just reviewed https://github.com/texastribune/donations/pull/835/commits/f2c6e45149c337a9a5d77d4ca37eff9679651266. Eventually we probably could make the `#stripe` channel its own env var, separate from `SLACK_CHANNEL`; we also use it [here](https://github.com/texastribune/donations/blob/9c73aaf24bef90194e2507ce8f2a786e48a28cf6/charges.py#L24) I noticed. I don't think that needs to happen here and there could be reasons we don't want to couple those just yet. 

Overall this seems like a simple and quick win for knowing when/if that runs. :shipit:

PR #842, opened 2022-01-10
Fix "failed to send email" for batch logs

ashley-hebler:
#### What's this PR do?

Adds encoding to the batch logs that send to the internal group,`stripe@texastribune.org`

Currently, the nightly log won't send if something like a letter with an accent mark is encountered. This causes headaches for folks that depend on reviewing this email.

This [SO post](https://stackoverflow.com/questions/54853197/the-smtplib-server-sendmail-function-in-python-raises-unicodeencodeerror-ascii/54853576#54853576) seems similar to what's happening:
> smtplib.server's sendmail method expects a bytes instance; if it gets a str it tries to encode it to ASCII, resulting in a UnicodeEncodeError if the str contains any non-ASCII characters.

I'm not actually sure. I was hoping we could just merge and...
1. Check in with the stripe@ group that this email send and is still readable
2. Errors similar to [this one](https://sentry.io/organizations/the-texas-tribune/issues/2886285476/?project=1304832&query=is%3Aunresolved&statsPeriod=30d) don't show up in Sentry again

The stripe@ group knows we're working on this

I don't really think this is the best solve because the places with the characters are still going to look like nonsense from what I understand. I thought this minimal fix could suffice since it has a small footprint and is pretty low priority because it isn't for a public audience 

[
"Look into failed to send mail: 'ascii' codec can't encode characters in donations"](https://3.basecamp.com/3098728/buckets/736178/todos/4395801568)

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/batch.py
+++ b/batch.py
@@ -39,7 +39,7 @@ class Log(object):
         """
         Send the assembled log out as an email.
         """
-        body = "\n".join(self.log)
+        body = "\n".join(self.log).encode("utf-8")
         recipient = ACCOUNTING_MAIL_RECIPIENT
         subject = "Batch run"
         send_email(body=body, recipient=recipient, subject=subject)


matthewdylan (approved):
LGTM!

PR #851, opened 2022-01-18
Better handle donor names with special characters

ashley-hebler:
#### What's this PR do?

During the batch charging process, we log the donor names for our own internal reporting. This will take that donor name and log it with special encoding that replaces special (aka non-ascii) characters with `?`

The entire email currently fails to send when a character like that is encountered.

_PR notes:_ I've been working through this issue in a series of PRs https://github.com/texastribune/donations/pull/842 https://github.com/texastribune/donations/pull/845 https://github.com/texastribune/donations/pull/847

It's hard to test but I ended up just using this sample snippet locally to double check how it'd print:

```py
class Log(object):
    def __init__(self):
        self.log = list()

    def it(self, string):
        self.log.append(string)

    def send(self):
        body = "\n".join(self.log)
        print(body)

log = Log()

log.it("---Starting batch job...")

# special char in O’
opportunities = ['Test O’Name', 'Test name 2']


log.it("---Processing charges...")

processing_msg = f"Found {len(opportunities)} opportunities available to process."
log.it(processing_msg)

for opportunity in opportunities:
  try:
      entry_name = opportunity
      encoded_name = entry_name.encode("ascii", "replace")
      decoded_name = encoded_name.decode("ascii")
      log.it(
          f"---- Charging $$$ to ({decoded_name})"
      )
  except:
      log.it(
          f"---- Charging $$$ to ({opportunity})"
      )
      print(f"Could not convert {opportunity} to ascii")

log.send()
```

I've been in contact with the email group who uses this log

It's probably a weird pattern to encode and then decode, but generally I was of the mindset of the simpler, the better for something only used internally.
We could also probably remove the try: except: once we confirm this works.

https://3.basecamp.com/3098728/buckets/736178/todos/4395801568

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/batch.py
+++ b/batch.py
@@ -105,9 +105,19 @@ def charge_cards():
         if not opportunity.stripe_customer:
             continue
         amount = amount_to_charge(opportunity)
-        log.it(
-            f"---- Charging ${amount} to {opportunity.stripe_customer} ({opportunity.name})"
-        )
+        try:
+            entry_name = opportunity.name
+            # replaces non-ascii characters with "?" - See PR #851
+            encoded_name = entry_name.encode("ascii", "replace")
+            decoded_name = encoded_name.decode("ascii")
+            log.it(
+                f"---- Charging ${amount} to {opportunity.stripe_customer} ({decoded_name})"
+            )
+        except:
+            log.it(
+                f"---- Charging ${amount} to {opportunity.stripe_customer} ({opportunity.name})"
+            )
+            logging.warn(f"Could not encode {opportunity.name}")
         try:
             charge(opportunity)
         except ChargeException as e:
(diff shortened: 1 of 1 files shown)


ashley-hebler:
@matthewdylan and @mailbackwards, I added you both as reviewers, but it can just be a whoever gets to it first kinda deal. This isn't a huge change, but just want to make this approach isn't _too_ strange. I went with the ascii + replace encoding bc utf-8 was harder to read. Ex: `Test O\xe2\x80\x99Name` vs `Test O?Name`

mailbackwards (approved):
Encoding confuses me so bad so I say we roll with this, and if even this doesn't work, we should do one or two things:
- set up a testing process for this batch script (whether automated or manual) that is more built into the codebase
- discuss another place to log/automate these with Anna and the accounting team that isn't email

PR #900, opened 2022-04-01
Add new tier of circle membership

ashley-hebler:
#### What's this PR do?

- Adds founders' circle to the circle page
- Also prevents wall group headers from showing up if no one is in that group yet. That's the case for this one because it's new.

This is a new tier set up in honor of the guys that started it all!

- Go to /cirlce
- Confirm that there's a new founders circle option in the form

I'll let Irma and Terry know

Yes, there is a [smelly template hack](https://github.com/texastribune/donations/pull/900/files#diff-74bdbdc8f9f246de9d1aad04dce97b58d8ce6a869e6f01ee9083e7be6dc6785aR9) in that CircleWall.vue file. You won't see the template change on the page until there are members of the founders' circle. My reasoning is I didn't want to refactor the wall template too drastically for this one-off, but if we start to have more of these we'll want to make that part of the group data.

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwxXBBX2S2vWpUiC/recsYFF5cWlKyY1Jd?blocks=hide

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/static/js/src/Wall.vue
+++ b/static/js/src/Wall.vue
@@ -8,20 +8,22 @@
         :class="{ 'grid_separator--l': groupIndex !== memberGroups.length - 1 }"
         class="wall__group"
       >
-        <h3 class="wall__heading grid_separator--s">{{ group.name }}</h3>
-        <slot name="subheading" :group="group"></slot>
-        <ul class="wall__list">
-          <li
-            v-for="(member, memberIndex) in group.members"
-            :key="memberIndex"
-            class="wall__item"
-          >
-            <span class="wall__star c-icon c-icon--yellow c-icon--baseline"
-              ><svg aria-hidden="true"><use xlink:href="#star"></use></svg
-            ></span>
-            <slot name="member" :member="member"></slot>
-          </li>
-        </ul>
+        <template v-if="group.members">
+          <h3 class="wall__heading grid_separator--s">{{ group.name }}</h3>
+          <slot name="subheading" :group="group"></slot>
+          <ul class="wall__list">
+            <li
+              v-for="(member, memberIndex) in group.members"
+              :key="memberIndex"
+              class="wall__item"
+            >
+              <span class="wall__star c-icon c-icon--yellow c-icon--baseline"
+                ><svg aria-hidden="true"><use xlink:href="#star"></use></svg
+              ></span>
+              <slot name="member" :member="member"></slot>
+            </li>
+          </ul>
+        </template>
       </section>
     </div>
   </section>
--- a/static/js/src/Wall.vue
+++ b/static/js/src/Wall.vue
@@ -8,21 +8,20 @@
         :class="{ 'grid_separator--l': groupIndex !== memberGroups.length - 1 }"
         class="wall__group"
       >
-        <template v-if="group.members">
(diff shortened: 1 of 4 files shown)


static/js/src/Wall.vue
```diff
@@ -8,19 +8,22 @@
...
-            :key="memberIndex"
-            class="wall__item"
-          >
-            <span class="wall__star c-icon c-icon--yellow c-icon--baseline"
-              ><svg aria-hidden="true"><use xlink:href="#star"></use></svg
-            ></span>
-            <slot name="member" :member="member"></slot>
-          </li>
-        </ul>
+        <template v-if="group.members">
```
ashley-hebler, line 11: the only change here is really the line above ^
`<template v-if="group.members">`
That's to prevent the new Founders' group from showing up until we have members in it

templates/circle-form.html
```diff
@@ -76,6 +76,11 @@ <h3 class="smallcaps--light grid_separator"><strong>Special Recognition</strong>
         </div>
       </section>
 
+      <div class="grid_container--l grid_separator--xl">
+        <!-- where the wall component attaches -->
+        <div id="circle-wall"></div>
+      </div>
+
```
ashley-hebler, line 83: Request from development team to bring the wall above the contact info

static/js/src/entry/circle/CircleWall.vue
```diff
@@ -5,6 +5,16 @@
       <figure class="wall__logo grid_separator--l" aria-hidden="true">
         <img src="./img/circle-logo.png" alt="Circle Membership logo" /></figure
     ></template>
+    <template #subheading="{ group }">
+      <template v-if="group.name === 'Founders\' Circle'"
```
SimmonsRitchie, line 9: Since "Founders' Circle" is also referenced in orderedGroupNames it might be worth considering whether to convert "Founders' Circle" into a constant. Down the line, I can imagine we could easily create a bug if we change the name of "Founders' Circle" and miss the other reference.

SimmonsRitchie (approved):
LGTM. I also tested it with some fake member data to make sure that founder members were showing up:

image

Agreed, the v-if with the founders' subheading is a bit of a smell, and, as you say, I'd be a bit tempted to do some refactoring, but yeah, it's probably overkill for now if we think it's unlikely we'll need subheadings for other circle tiers.

static/js/src/entry/circle/CircleWall.vue
ashley-hebler: GREAT point. I ended up doing this: [097e7eb](https://github.com/texastribune/donations/pull/900/commits/097e7eb3886d996ab90e4131f832bbab8eaabe77) to address that. I thought about doing something like [`orderedGroupNames[0]`](https://github.com/texastribune/donations/blob/097e7eb3886d996ab90e4131f832bbab8eaabe77/static/js/src/entry/circle/CircleWall.vue#L35) so the label lived in on place, but we might change the order and forget. It's at least closer to where you'd update and more greppable now, but still redundant. Seems better than what I originally had nevertheless! Thanks for flagging.

ashley-hebler:
Thanks for the review @SimmonsRitchie! LOL I love your fake data. 🤞that MJ sees this.

PR #918, opened 2022-05-06
turn off thermometer and uncomment messaging area

matthewdylan:
for end of smd 22

#### What's this PR do?
Turns off the thermometer on the support page and removes the timely messaging block tied to membership drive.

SMD 22 is over.

make
make restart
visit localhost/donate and ensure the thermometer and the messaging block are gone

Don't think it needs to be, but we can announce in general if we need to.

Previously the scss for the messaging block had been removed when tearing down after a support event, but I'm leaving it in this time around so there's less to think about next md (and it seems relatively harmless and will be used again in the future). That said, happy to hear any counter thoughts.

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwCUOttsruMqY7r7/rec2TKhmvv4Xrijqk?blocks=hide

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/app.py
+++ b/app.py
@@ -58,7 +58,7 @@ from util import (
 )
 
 ZONE = timezone(TIMEZONE)
-USE_THERMOMETER = True
+USE_THERMOMETER = False
 
 if ENABLE_SENTRY:
     import sentry_sdk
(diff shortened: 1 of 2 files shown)


ashley-hebler (approved):
LGTM! Good call to leave the message container in there 😌

PR #922, opened 2022-05-13
Malformed Salesforce IDs

djpeacher:
#### What's this PR do?
This PR removes the donation form validation for the Salesforce campaign and referral query parameters.

If a bad Salesforce ID is provided, specifically if the length was greater than 18, the entire donation would fail. We shouldn't have to worry about bad IDs in the form because there is already logic in place to remove those IDs during API calls to Salesforce if they are invalid:
https://github.com/texastribune/donations/blob/a2dd48da636cefaef4096ced1a820040be795f44/npsp.py#L513-L529

- This was the original URL for this ticket, it should now allow a donation: http://local.texastribune.org/donate?installmentPeriod=once&amount=100&campaignId=7016f0000023MsvAAE%3Dtrib-ads-paid&utm_campaign=trib-marketing&utm_term=smd
- Any random/invalid `campaignId` should still let the donation process. (You will see an error/retry log in your terminal)
  - http://local.texastribune.org/donate?campaignId=ERROR
  - http://local.texastribune.org/donate?campaignId=ERROR_WITH_LONG_STRING
- Any random/invalid `referralId` should still let the donation process.  (You will see an error/retry log in your terminal)
  - http://local.texastribune.org/donate?referralId=ERROR
  - http://local.texastribune.org/donate?referralId=ERROR_WITH_LONG_STRING
- A valid `campaignId` and `referralId` should also allow the donation to process. I don't have an example for a referral, but here is an example for a campaign courtesy of Ashely!
  - http://local.texastribune.org/donate?campaignId=70117000001umA6

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwgnCRCqnRWlO1In/recKQOPNIaEnAEJSW?blocks=hide

### DIFF
--- a/forms.py
+++ b/forms.py
@@ -67,8 +67,8 @@ class BaseForm(FlaskForm):
         u"Pay Fees Value", [validators.AnyOf(["True", "False"])]
     )
     reason = StringField(u"Encouraged to give by", [validators.Length(max=255)])
-    campaign_id = HiddenField("Campaign ID", [validators.Length(max=18)])
-    referral_id = HiddenField("Referral ID", [validators.Length(max=18)])
+    campaign_id = HiddenField("Campaign ID")
+    referral_id = HiddenField("Referral ID")
 
 
 class DonateForm(BaseForm):


ashley-hebler (approved):
Test links WFM and good sleuthing to find that the try/except already has that campaign logic factored in. I do kinda wonder if having the maxlength is beneficial for some security, sanitation reason I'm not thinking of. I noticed [we don't validate length](https://github.com/texastribune/donations/blob/master/forms.py#L117-L118) on the Blast form currently so it matches that again.

I'll follow up with @x110dc about this, in case I'm missing something, but I say good to merge to fix this edge case failed donation scenario. Thanks @djpeacher 

Oh and deployment note on this repo is that we have to deploy through Heroku CLI right now. We can offline about that.

djpeacher:
> I noticed [we don't validate length](https://github.com/texastribune/donations/blob/master/forms.py#L117-L118) on the Blast form currently so it matches that again.

Ah yes, that was another justification I found for removing the validation.

PR #973, opened 2022-10-06
add credit card update feature

matthewdylan:
#### What's this PR do?
Adds functionality to allow users to update their credit card info in the user management portal.

Members have long asked for this ability and it undoubtedly makes the member experience smoother.

Testing steps can be found [here](https://docs.google.com/document/d/1QYo7syZ_Y31hbEZa4vaBVwdp04wvZ7EFOyfpSXjrBwE/edit).

We're in close communication with the membership team on this projects status, but for users/members, this will be a soft launch.

#### TODOs / next steps:

* [ ] *your TODO here*


static/js/src/entry/account/routes/membership/components/RecurringOrCircle.vue
```diff
@@ -1,13 +1,37 @@
...
+            <button @click="openPaymentForm = !openPaymentForm">
+              <icon v-if="!openPaymentForm" name="pencil-fill" :display="{ size: 's' }" />
+              <icon v-if="openPaymentForm" name="close" :display="{ size: 's' }" />
+            </button>
+            <card-update
+              v-if="openPaymentForm"
+              @formSubmitted="formSubmitted"
+              @onSuccess="onSuccess"
+              @badCard="badCard"
+            ></card-update>
```
SimmonsRitchie, line 34: Style-wise, I think the pencil icon looks slick, I like how it transforms into an X on click. But I have two concerns:

A) I feel it looks good when it's on the same line, but looks a bit weirder when it falls below it. I think if we go with the icon, it might be best to have it all on one line.

Screen Shot 2022-10-11 at 11 58 12 AM

B) Maybe it's just me, but part of me worries more generally that clicking on the pencil icon to edit credit card details may not be intuitive to older or less tech-saavy users. I'd hate to lose the icon, but I do wonder if this is a case where we need to be a bit more explicit. We could replace the icon with text or a button that simply reads "Edit" or "Edit details". A compromise might to include "Edit" with the icon. For example, something like:

Screen Shot 2022-10-11 at 12 18 16 PM

ashley-hebler: +1! I think the word + icon is most intuitive. Great idea @SimmonsRitchie

SimmonsRitchie:
I know this has been a long-requested feature and it's great to see it up and running! All test steps WFM. Looking over the JS, everything looks pretty clean. Although I can take another pass later.

I left a note regarding the pencil icon. Two other notes:

1) I hit an error while trying to update my zip code. Despite that, the app appears to have actually saved my new zip code correctly. I have a suspicion this might be unrelated to this PR however and just be one of the general quirks of this app, especially when run locally, but I'm flagging it here just in case.

Screen Shot 2022-10-11 at 11 55 00 AM

2) From a UI perspective, there are certain mobile/tablet breakpoints where the credit card text is a bit jumbled:

Screen Shot 2022-10-11 at 1 06 37 PM

It's not a dealbreaker, it's still useable, but I wonder if we could make this a little clearer by simply tweaking the responsive settings for this component. Currently, for viewports smaller than 520px ($bp-s in Queso) these three sections stack as a column. Perhaps if we tweaked the CSS to stack for screens that are less than 845px ($bp-l) we'd be sorted. You might have to play around with it.

ashley-hebler (commented):
Test steps and functionality all LGTM. The implementation seems nice and clean, but I imagine was a lot to sort through at first; cheers to you on that! This is such a great and valuable feature.

I left some very small inline notes, but I'll drop two general thoughts here too:

_Logs/warnings/tracking_
It would be good to have some way to see when people are using this and if they're encountering errors. The errors part is probably more on the API side or maybe just built-in, but just want to confirm we have a way of knowing if the form glitches on users sometimes. 
For the better scenario of the update successfully running, could we add some GA trackers maybe, similar to the [profile update form](https://github.com/texastribune/donations/blob/9e95dca948b8654e10991a6948d8fb9abf66b8ee/static/js/src/entry/account/routes/edit-contact-info/containers/EditFormContainer.vue#L145-L190)? It's a weird suggestion since we'll have to rewrite event tracking for GA4, but in the early days of roll-out, I think it'd be helpful to see usage. I think the GA tracking is a nice to have so if that ends up looking like a big undertaking, we can shelve for post-launch. 

_UI bug Dan pointed out_
I think stacking the columns is a beautiful (and quick!) solve for the form jumbled text issue. I'd even say, it'd be fine if they were always stacked, avoiding all the need for breakpoints and factoring in when the sidebar is expanded. In that case we might not need that to be a [`c-info-list`](https://texastribune.github.io/queso-ui/sections/components/c-info-list/) at all. The design team has a ticket for the UI portion of this so the simplest fix here makes the most sense since we may adjust it later.

TL;DR
As long as we can see errors, I say this is good to go 🚀 🤑! Agree with Dan that the form ui bug isn't a real blocker, but would be nice to fix if it's not a pain.

static/js/src/entry/account/routes/membership/components/RecurringOrCircle.vue
matthewdylan: Yeah, good call on this @SimmonsRitchie. Done and done.

static/svg/pencil-fill.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-pencil-fill" viewBox="0 0 16 16"><path d="M12.854.146a.5.5 0 0 0-.707 0L10.5 1.793 14.207 5.5l1.647-1.646a.5.5 0 0 0 0-.708l-3-3zm.646 6.061L9.793 2.5 3.293 9H3.5a.5.5 0 0 1 .5.5v.5h.5a.5.5 0 0 1 .5.5v.5h.5a.5.5 0 0 1 .5.5v.5h.5a.5.5 0 0 1 .5.5v.207l6.5-6.5zm-7.468 7.468A.5.5 0 0 1 6 13.5V13h-.5a.5.5 0 0 1-.5-.5V12h-.5a.5.5 0 0 1-.5-.5V11h-.5a.5.5 0 0 1-.5-.5V10h-.5a.499.499 0 0 1-.175-.032l-.179.178a.5.5 0 0 0-.11.168l-2 5a.5.5 0 0 0 .65.65l5-2a.5.5 0 0 0 .168-.11l.178-.178z"/></svg>
```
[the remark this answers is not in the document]
matthewdylan: I ran it through here and maybe I don't know what I'm doing, but it made very little difference (you can see what it looks like now), so let me know if I missed something.

static/js/src/entry/account/routes/membership/components/RecurringOrCircle.vue
```diff
@@ -1,13 +1,37 @@
 <template>
   <section class="c-detail-box">
     <div class="has-xxl-btm-marg">
+      <p
+        v-if="declinedCard"
+        role="alert"
+        class="has-b-btm-marg has-text-error"
+      >
+        <strong>Card was declined</strong>
+      </p>
```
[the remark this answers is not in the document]
matthewdylan: I need to think through portal-api being down or hiccuping... think that would be applicable to the profile update form as well, as you mentioned, so that can PROBABLY be considered outside of this project BUT I did make some updates to error handling in the case where portal-api doesn't get the exact pieces it's currently expecting to get, which should clear up some of that "Wait, why is my card reading as declined?"

matthewdylan:
Thanks to you both for your suggestions here. Think I followed up on all of these and they all definitely make the feature better in various ways, so much appreesh for that.

PR #982, opened 2022-11-15
Fixed Donation Fee Rounding Bug

djpeacher:
#### What's this PR do?
Frontend JS now calls `round()` on fees instead of `floor()`.

The backend was rounding up/down, but the frontend was only rounding down. This caused the frontend to display some fees off by 1 cent.

- Enter $20 on production, and you will see a fee of $0.75 on the frontend.
- Enter $20 locally, and you will see a fee of $0.76 on the frontend.
  - If you perform the donation (with fee), you can confirm Salesforce shows $20.76 as the total.
- You can compare frontend and backend calculation:
  https://github.com/texastribune/donations/blob/f5fcc5b8f9c27845d1d5ac353de476d821291c75/static/js/src/connected-elements/PayFees.vue#L53-L54
  https://github.com/texastribune/donations/blob/f5fcc5b8f9c27845d1d5ac353de476d821291c75/charges.py#L44-L52

- Inform membership team.

- There is still some question about what the actual fee is that Stripe is taking, but this codebase would not be affected by a fee change on their end. We are hard coded to a particular rate.

- https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwvu5Kdv9ASOAfh2/recX7kzowQu3TYdWj?blocks=hide

### DIFF
--- a/static/js/all.js
+++ b/static/js/all.js
@@ -91,7 +91,7 @@ var payFeeAmount = function() {
     payFeeElement.text('');
   } else {
     var totalAmount = (goalAmount + .30) / (1 - 0.022);
-    var feeAmount = Math.floor((totalAmount - goalAmount) * 100) / 100;
+    var feeAmount = Math.round((totalAmount - goalAmount) * 100) / 100;
 
     payFeeElement.text('$' + feeAmount.toFixed(2));
   }
--- a/static/js/blast.js
+++ b/static/js/blast.js
@@ -35,7 +35,7 @@ var listen_for_installments = function() {
 var payFeeAmount = function() {
   var goalAmount = parseFloat($('input[name="amount"]:checked').val()),
       totalAmount = (goalAmount + .30) / (1 - 0.022);
-      feeAmount = Math.floor((totalAmount - goalAmount) * 100) / 100,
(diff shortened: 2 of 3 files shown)


static/js/all.js
```diff
@@ -91,7 +91,7 @@ var payFeeAmount = function() {
     payFeeElement.text('');
   } else {
     var totalAmount = (goalAmount + .30) / (1 - 0.022);
-    var feeAmount = Math.floor((totalAmount - goalAmount) * 100) / 100;
+    var feeAmount = Math.round((totalAmount - goalAmount) * 100) / 100;
```
ashley-hebler, line 94: Small thing but for clarity we could mention that the `feeAmount` equation rounds to two decimals. That formula made me have to stop and think and a comment could maybe quickly convey what's that's doing.

ashley-hebler (approved):
Test steps WFM and your logic makes sense here. Excellent find!

static/js/all.js
djpeacher: Great idea 💡 It also took me a sec to realize what this line was doing!

ashley-hebler: Nice like to the stripe docs too! Thanks @djpeacher

PR #1041, opened 2023-04-05
Blast Meta 💥 

djpeacher:
#### What's this PR do?
- Adds custom meta tags for the blast. Specifically updates the images and description.
- BONUS: Updated title and added twitter `summary_large_image` tag.

- Better social share experience for the blast!

- Can't really manually test, so just validate the markup.

- #team-mar-com

- Does this also need to appear on https://support.texastribune.org/blast-promo?
  - Also kinda funny to me that there's a public link with a permanent promo, I think, but I just might be missing some context.

- https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwULM503frwBhAOe/recalhswlRZ2FqOVD?blocks=hide


templates/blast-form.html
```diff
@@ -6,6 +6,15 @@
 <script src="{{ url_for('static', filename='js/blast.js') }}"></script>
 {% endblock %}
 
+{% block og_meta %}
+  <meta property="og:url" content="https://support.texastribune.org/blastform">
+  <meta property="og:image" content="https://static.texastribune.org/media/membership/tt-blast-share.png">
```
ashley-hebler, line 11: **question:** Would it make sense to store this at the repo level?  We seem to do that for the social image [on the main page](https://github.com/texastribune/donations/blob/ad53ecc8f4d031c800b23b4ca6ccd7d786933f78/templates/new_layout.html#L15).

templates/blast-form.html
```diff
@@ -6,6 +6,15 @@
...
+  <meta property="og:title" content="The Blast | The Texas Tribune">
+  <meta property="og:type" content="website">
+  <meta property="og:description" content="The Blast, The Texas Tribune’s premier newsletter, puts to use the full weight of the newsroom’s political team to serve up exclusive reporting, nonpartisan analysis and the first word on happenings in the Capitol three times a week for insiders. Don’t get left behind. Subscribe today.">
+  <meta name="twitter:card" content="summary_large_image">
```
ashley-hebler, line 15: **praise:** Good call! I bet they prefer this so good instincts

ashley-hebler (approved):
LGTM!

> Also kinda funny to me that there's a public link with a permanent promo

I had the same thought in reviewing this. I don't remember if there's a backstory. Mind checking in with Natalie C. on this? Maybe we just 🪓 that route.

templates/blast-form.html
djpeacher: Ah didn't notice that, great suggestion! :done:

djpeacher:
Thanks for the review @ashley-hebler! Natalie said to keep the blast promo route and have the meta match the main blast route 🚀 💥

PR #1049, opened 2023-04-24
Move core app files to /server folder

matthewdylan:
#### What's this PR do?
* Adds /server and /tests folders to project structure
* copies the core .py files from root into the /server folder and updates the imports therein
* copies tests.py from root into /tests folder (w/ a temp rename) and updates the imports therein
* updates circle.sh to use the same testing command as docker-compose.yml

This a piece of a larger initiative to move closer to a project structure meant for a medium-sized project (as seen [here](https://flask.palletsprojects.com/en/2.2.x/patterns/packages/#simple-packages)). This pr is only meant to build out the new project structure and hook up the testing commands to work using said structure.

The only remaining pr will be one that creates a new make command that mimics the functionality of the current 'interactive' command but simplifies it using the docker-compose file. This pr will also handle the moving of the /templates and /static folders to the /server folder (and rejiggering references to account for this).

As mentioned in the comments below, there is a longer term desire to move the Vue pieces out of the /static folder and into the root folder, but given the scope creep already (of my own doing, admittedly) I'm going to kick this move down the road to the forthcoming vue 3 upgrade.

run make test and ensure all tests pass

N/A

Besides the new folders and the updated imports, this pr is largely just a copy job so that further changes are easier to see and don't get lost in moving files around. It should be noted that this isn't wiring up any of the frontend to work with these new setup, but that is forthcoming in the next pr.

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/recm2EZHmnXTRlB2b?blocks=hide

#### TODOs / next steps:

* [ ] *your TODO here*


.gitignore
```diff
@@ -24,3 +24,4 @@ npm-debug.log
 .mypy_cache/
 .vscode/
 cov.html
+**.DS_Store
```
ashley-hebler, line 27: **note:** [In other Tribune repos](https://github.com/texastribune/texastribune/blob/master/.gitignore#L1-L5) we tend to do this individually at a local level using `.git/info/exclude` since it's not technically a file we'd all share. That said, it does still throw off new engineers (fellows especially)that are setting up their machines for the first time. TL;DR I don't mind having it here!

ashley-hebler (commented):
@matthewdylan thanks for taking the time to break this apart! It's helpful to see these incremental changes.

This is purely an opinion, but my initial thought is that I'd expect to see front-end files in a directory called `src`. For this use I'd think to call the folder `app` or something. Then again, I'm not really sure what the overall file structure of the app is going to look like. Could you drop that in the comments or PR notes?

.gitignore
matthewdylan: Oh yeah good point... that's a goof on my part... I haven't had to mess with the .DS_Store since I first got here and I couldn't remember how I handled it then, but this jogged my memory on that. I wonder why they started showing up again for me... either way, not really relevant. I'll remove from here and handle it locally. Thanks for the reminder.

matthewdylan:
> @matthewdylan thanks for taking the time to break this apart! It's helpful to see these incremental changes.
> 
> This is purely an opinion, but my initial thought is that I'd expect to see front-end files in a directory called `src`. For this use I'd think to call the folder `app` or something. Then again, I'm not really sure what the overall file structure of the app is going to look like. Could you drop that in the comments or PR notes?

I think I was influenced by some of the core flask libraries since I've been digging more into flask lately and they use 'src' as the main app folder in their projects ([as seen here](https://github.com/pallets/flask)). That said, I think we do ultimately want to move the vue piece of this out to the root level, and I'm seeing the app->python framework, src->javascript framework setup over at least a handful of projects I've searched out, so I'm leaning in your direction for that reason. A few other ideas worth throwing around... django would usually suggest the 'app' folder have the same name as the project overall, so in this case "donations" (flask recommends this too, fwiw [as seen here](https://flask.palletsprojects.com/en/2.3.x/patterns/packages/)). I don't love this, personally, because I think it's a little confusing to have the same name for the project and app (thus using src in this case). I also saw use of server->python framework, client->javascript framework, which I like, because I think it's a little more descriptive than app/src, but it is also less common, so it might not be as immediately familiar to someone new to the project. I'll leave some notes on what the remaining moves here are in the pr notes.

.gitignore
djpeacher: I actually don't think I knew this 📝

matthewdylan:
> Hey @matthewdylan, sorry for taking a bit to get to this one!
> 
> Overall, the changes LGTM and make sense. Yay for `pytest`! I also put in my votes for possible folder structure naming in Slack!
> 
> And just to confirm, I couldn't get the site to spin up here, but that should be expected in this intermediate PR right?

That's right, the next part has the app working as it's supposed to. Thanks for looking @djpeacher!

matthewdylan:
> > @matthewdylan thanks for taking the time to break this apart! It's helpful to see these incremental changes.
> > This is purely an opinion, but my initial thought is that I'd expect to see front-end files in a directory called `src`. For this use I'd think to call the folder `app` or something. Then again, I'm not really sure what the overall file structure of the app is going to look like. Could you drop that in the comments or PR notes?
> 
> I think I was influenced by some of the core flask libraries since I've been digging more into flask lately and they use 'src' as the main app folder in their projects ([as seen here](https://github.com/pallets/flask)). That said, I think we do ultimately want to move the vue piece of this out to the root level, and I'm seeing the app->python framework, src->javascript framework setup over at least a handful of projects I've searched out, so I'm leaning in your direction for that reason. A few other ideas worth throwing around... django would usually suggest the 'app' folder have the same name as the project overall, so in this case "donations" (flask recommends this too, fwiw [as seen here](https://flask.palletsprojects.com/en/2.3.x/patterns/packages/)). I don't love this, personally, because I think it's a little confusing to have the same name for the project and app (thus using src in this case). I also saw use of server->python framework, client->javascript framework, which I like, because I think it's a little more descriptive than app/src, but it is also less common, so it might not be as immediately familiar to someone new to the project. I'll leave some notes on what the remaining moves here are in the pr notes.

Follow up here that after polling the group, there was a majority who agreed that "server" was a better solution here than "src". The pr has been updated accordingly.

PR #1067, opened 2023-05-19
Reorg project structure

matthewdylan:
#### What's this PR do?
Reorganizes the donations app to...
* follow the "simple package" project structure mentioned [here](https://flask.palletsprojects.com/en/2.3.x/patterns/packages/#simple-packages)
* utilize docker-compose to clean up the Makefile

Also ...
* upgrades flask to 2.2.3
* pins requests to 2.29.0 because of a forced upgrade to urllib3 2.0 in 2.30.0 that was breaking our things

* makes spinning the app up a simpler process
* makes writing tests easier and more modular
* makes the app pieces more modular as well
* adheres closer to our project structure in the texastribune repo, adding consistency
* hopefully this all adds up to a simpler flow for engineers when needing to update or add functionality to the app

You can test locally or on [test](https://donations-testing.texastribune.org/donate)

If testing locally, first:
* ensure your local env-docker file has the const `SLACK_CHANNEL` set to `#tech-test`
* run `make`

Either way:
* test making a donation (should show up in slack -> #tech-test if testing locally and #stripe if on test)

Shouldn't need to be as the updates here are structural.

Now that our project structure is more advanced, there is more work we can do here like...
* move vue pieces out of /server/static/js and into their own folder at root ('client' or 'frontend')
* eventually modularize the app.py file by moving the routes to a views.py file and consider moving some of the other functions to utils or elsewhere, as appropriate

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/recm2EZHmnXTRlB2b?blocks=hide

* [x] Updated the documentation/wiki? *( )*

#### TODOs / next steps:

* deploy to production


ashley-hebler (approved):
All WFM! I took this on one last test drive. Couldn't find any launch blockers!

Small notes/observations (non-blocking):
- Every so often I'm seeing a `'NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x400ebc5790>: Failed to establish a new connection: [Errno 110] Connection timed out')'` line in my terminal. Very possible that was happening before but just wanted to note. I don't see it in staging logs which is a good sign!
- I noticed we're bumping Flask in this process too, which should squash a vulnerability 🎉
     - Could we mention that detail in the PR note. It'd be easy to miss with the large diff
     - Similarly we could also mention adding requests
- Should we create a git tag and associated release _before_ we merge this? I know we don't use those, but might be nice to have a quick bookmark of sorts to see the setup pre-reorg.
- Thank you so much for breaking this up in a way to keep most of the git history clean. I'm sure there are commits in the app.py file that are a little spread out, but if we wanted to track stuff back through history, we should be able to. Truly appreciate the care you put into that!

requirements.txt
```diff
@@ -1,4 +1,6 @@
-Flask==2.0.2
+Flask==2.2.3
+# pinning requests because 2.30.0 is breaking for now
+requests==2.29.0
```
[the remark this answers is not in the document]
matthewdylan: haha... good question... admittedly I also forgot exactly why this is breaking, only that it is and in the moment I was being lazy... let me track down our dialogue and see if I can link to some kind of explainer here

docker-compose.test.yml
bfreeds, line 1: **praise:**  Nice usage of the docker-compose override pattern 🐳

docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+
+  rabbitmq:
+    image: rabbitmq:management
+    ports:
+      - target: 5672
+        published: 5672
+        protocol: tcp
+        mode: host
+      - target: 15672
+        published: 15672
```
bfreeds, line 38: suggestion(syntax):  I think you could use the _short-form_ syntax here, which would be a little easier to read & be consistent above with the port declaration for the `dev`/flask container.

https://docs.docker.com/compose/compose-file/compose-file-v3/#short-syntax-1

```suggestion
      - 5672:5672
      - 15672:15672
```

docker-compose.yml
bfreeds, line 1: **idea**:  In the future, we might want to / could configure this docker-compose to use https, to be closer to prod.

When running `make dev`, I noticed stripe errors seemingly related to needing https.  The app seems to work for me locally, though!
image

bfreeds (approved):
@matthewdylan this looks so good!  This looks like it took a lot of work, but is going to make this app so much easier to work with, read, and understand (will save our future selves lots of time and energy).  Thank you for doing this.

I reviewed and mostly looked at your docker-compose setup (so nice to have here).  I added a couple small notes, but nothing blocking merge.

Thank you for such a thorough and thoughtful PR.  This note was especially useful/appreciated 😅:
> ensure your local env-docker file has the const SLACK_CHANNEL set to #tech-test

Dockerfile
```diff
@@ -26,4 +27,6 @@ RUN pip install -r /app/dev-requirements.txt
 
 COPY . /app
 
+RUN pip install -e .
```
djpeacher, line 30: **question:** I probably missed this in a previous PR, but what is this third pip install for?

docker-compose.yml
djpeacher, line 1: **praise:** These new docker-compose files are awesome! I love how simple startup and tests are now!

djpeacher (approved):
Hey @matthewdylan, this LGTM! I'm not seeing any obvious issues, tests ran successfully, and I was able to make a donation! 🚀 Loving this refactor!

docker-compose.test.yml
matthewdylan: thank you!

docker-compose.yml
matthewdylan: good call, updated accordingly

Dockerfile
matthewdylan: this installs the server package locally (seen in setup.py)

docker-compose.yml
matthewdylan: thank you!

matthewdylan: Okay, good note. I'll make an airtable task to check back in on this.

matthewdylan:
Thanks for all the reviews everybody! @ashley-hebler I'm wondering if that NewConnectionError is tied to the point @bfreeds made about HTTPS connections, but I'll dig in further either way (and keep an eye on the prod logs too).

PR #1095, opened 2023-08-10
hot fix to allow resending of stripe events

matthewdylan:
use the invoice created date when trying to find an opportunity instead of assuming it's always today... this is helpful if we ever need to rerun stripe events days after the charge went through


ashley-hebler (approved):
@matthewdylan and I slacked on this and we both feel good about shipping it to patch a Salesforce bug :shipit:

PR #1097, opened 2023-08-14
hot fix to get timezone accurate invoice timestamp

matthewdylan:
fromtimestamp keeps the original timezone as expected whereas using utcfromtimestamp is causing some fluctuations

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -1521,7 +1521,7 @@ def update_next_opportunity(opps=[], invoice=None):
             stage_name="Pledged", stripe_subscription_id=invoice["subscription"]["id"]
         )
 
-    charged_on = datetime.utcfromtimestamp(invoice["created"]).strftime('%Y-%m-%d')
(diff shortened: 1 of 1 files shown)


matthewdylan:
For more context... datetime.utcfromtimestamp() takes a unix timestamp and returns a readable/formattable instance in Coordinated Universal Time while datetime.fromtimestamp() takes a unix timestamp and returns a readable/formattable instance in the local timezone. My mistake was assuming the timestamp saved in salesforce for RDOs and opportunities was UTC when it was not. Ergo, with most instances, the utcfromtimestamp return was hitting the same date as what was saved in salesforce, but instances within an hour of two of the next day was returning the next day while salesforce opps had been saved with the current day. Because we are relying on this timestamp to grab the proper opportunity to update after a successful stripe charge, this utc/local timezone disparity led to not finding the opportunity in question in cases like mentioned above.

PR #1119, opened 2023-10-13
📧 Business Email Unicode Fix

djpeacher:
#### What's this PR do?
- Unicode characters are now stripped out of new business membership emails.

- Some business input unicode characters (specifically in the "organization name" field), which causes the email to fail to send.

- Make a /business donation (on staging) with a unicode character in the "organization name". See example:
Screenshot 2023-10-13 at 12 22 35 PM
- View Mailtrap via heroku. You should see the email, and the unicode character should have been stripped out.

- https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwULM503frwBhAOe/recIiPjXi5A3YNiWS?blocks=hide

### DIFF
--- a/server/util.py
+++ b/server/util.py
@@ -188,7 +188,7 @@ def send_email(recipient, subject, body, sender=None):
 
     TO = recipient if type(recipient) is list else [recipient]
     SUBJECT = subject
-    TEXT = body
+    TEXT = body.encode('ascii', errors='ignore').decode('ascii')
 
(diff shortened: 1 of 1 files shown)


ashley-hebler (approved):
LGTM! I wasn't able to really see what that body of the email turned out to be in Mailtrap but the code change makes sense so I feel good about it. Thanks for doing this!

PR #1120, opened 2023-10-13
Add donation type fallback

matthewdylan:
#### What's this PR do?
Falls back to a donation type getter based off of a subscription's product when no donation type is provided. Donation type is provided automatically from all of our pages, so this only applies to donations created through the stripe dashboard.

Currently stripe dashboard donations can't make it through to salesforce because they don't pass this info on through metadata like other donations do.

* In stripe, make sure you're in Test mode (upper right hand corner)
* Click the plus button (upper right hand corner) and choose "Subscription"
    * Screen Shot 2023-10-13 at 2 47 20 PM
* Click in the "Customer" field and choose any of the options (likely me)
* Click in the "Product" field and choose any circle product (founders, editors, chairmans, leadership)
* Lastly, towards the bottom of the form, make sure a payment form is chosen
    * you'll see the last 4 digits (likely 4242)
* Click "Create Test Subscription" at the bottom of the form
* You should see the donation come through on #tech-test in slack

I'll reach out to membership and Anna.

There are many possible solutions to this problem and I'd like to see if I could find a cleaner one once we update the front-end and can rethink our handling of product ids at that point.

Also worth considering if there is a place for admins to do this kind of donation add within ump at some point. Circle members in particular seem to pop up in this scenario, both because they ask admins to handle it, and because circle memberships aren't set it/forget it since they end after 3 years.


server/constants.py
```diff
@@ -0,0 +1,24 @@
...
+    "leadershipMonthly",
+    "leadershipYearly",
+    "editorMonthly",
+    "editorYearly", 
+]
+
+BLAST_PRODUCTS = [
+    "blast",
+    "blastTaxExempt",
+]
```
ashley-hebler, line 24: **thought:** I think you speak to this a little in the PR notes, but does this sorta create two sources of truth for these names? Or make it so that we're maintaining them in two places I mean. Not sure of a way around that but it just feels a little funny.

I'm referring to these other instances of them:
https://github.com/texastribune/donations/blob/bcb441aef4b165e80a1761ce3894d9c4c66eeb62/server/config.py#L66-L84 right?

ashley-hebler (commented):
Thanks for quickly taking action on this. I know it only started showing up on Friday.

So I _think_ the test steps worked for me with the exception of it showing up in Slack. I created two Blast subs and circle monthly from the dash which seems to reflect properly and then was trying to figure out how we'd create a regular sustaining. I didn't see that in the stripe dropdown and proceeded to create a product. I think that's wrong tho so feel free to clean it up. The entry is [here](https://dashboard.stripe.com/test/customers/cus_O4QwAi2bgvd1EF).

I was ultimately trying to figure out how you'd end up [in the final `else:`](https://github.com/texastribune/donations/pull/1120/files#diff-f624bf5d393b6db077aa91494d3cbbee702f78945bd2ffcf0b7ed98a3f339057R256) of the of the `type_getter` but I'm sure it's a lack of stripe savviness on my end. I left small nitpicky thought in the inline notes and let me know if that'd be easier to discuss in the open about.

server/constants.py
matthewdylan: mmm, good point... I made some updates to the STRIPE_PRODUCTS dict to simplify this a bit... let me know what you think

ashley-hebler: ah yeah that makes sense to me! I guess that means you'll have to refactor how we're calling that in a few other places too right? I'm seeing [this error](https://the-texas-tribune.sentry.io/issues/4554433156/?alert_rule_id=1608736&alert_type=issue&notification_uuid=ed6f399a-715a-4b80-8bb9-cdefe1b793e0&project=1304832) when I submit using the form.

matthewdylan: Alright alright alright, I ended up making this much simpler and moving the donation type and the donor_amount (basically the base amount sans fees) in stripe. I think it makes sense to make that the source of truth for that data, but let me know what you think @ashley-hebler.

ashley-hebler: oh ok yeah I think I follow and that sounds even better!

PR #1142, opened 2023-12-04
adding skip_sync logic

matthewdylan:
#### What's this PR do?
This adds a small chunk of code that keeps us from syncing anything to salesforce from a stripe event (like customer_subscription_created) when the field "skip_sync" is found on the subscription metadata.

We add the "skip_sync" field to the subscription metadata when converting salesforce recurring donations to stripe subscriptions since the info already exists in salesforce. We also update the recurring donation with the stripe subscription id (the only thing we need to update in salesforce) immediately after creating the subscription, so we don't need to worry about listening for the stripe event. Overall, this small bit of logic helps us to not duplicate recurring donations in salesforce.

Once the conversions from salesforce to stripe are done, we can probably remove this code. I can't think of a situation where a donation shouldn't be synced to salesforce except for this one

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/recJsJbQNe6QFgNMw?blocks=hide

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -868,15 +868,15 @@ def customer_subscription_created(event):
     # It starts by looking for a matching Contact (or creating one).
     subscription_id = event["data"]["object"]["id"]
     subscription = stripe.Subscription.retrieve(subscription_id, expand=["latest_invoice"])
-    sub_meta = subscription["metadata"]
+    subscription_meta = subscription["metadata"]
 
     # When migrating existing recurring donations from salesforce to stripe subscriptions, we pass a field
     # called "skip_sync" to the subscription metadata to let us know that we don't need to push anything back
     # to salesforce. In that instance, we exit the function at this point.
-    if sub_meta.get("skip_sync", False):
+    if subscription_meta.get("skip_sync", False):
         return None
 
-    donation_type = sub_meta.get("donation_type", subscription["plan"]["metadata"].get("type", "membership"))
+    donation_type = subscription_meta.get("donation_type", subscription["plan"]["metadata"].get("type", "membership"))
 
     invoice = subscription["latest_invoice"]
     invoice_status = invoice["status"]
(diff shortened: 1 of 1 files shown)


server/app.py
```diff
@@ -868,7 +868,15 @@ def customer_subscription_created(event):
     # It starts by looking for a matching Contact (or creating one).
     subscription_id = event["data"]["object"]["id"]
     subscription = stripe.Subscription.retrieve(subscription_id, expand=["latest_invoice"])
-    donation_type = subscription["metadata"].get("donation_type", subscription["plan"]["metadata"].get("type", "membership"))
+    sub_meta = subscription["metadata"]
```
ashley-hebler, line 871: **suggestion (non-blocking):** For readability/developer experience, it'd be more clear to me if this were `subscription_meta` to be consistent with the lines above. That could also be that I'm just jumping into this tho!

server/app.py
```diff
@@ -868,7 +868,15 @@ def customer_subscription_created(event):
...
+
+    # When migrating existing recurring donations from salesforce to stripe subscriptions, we pass a field
+    # called "skip_sync" to the subscription metadata to let us know that we don't need to push anything back
+    # to salesforce. In that instance, we exit the function at this point.
+    if sub_meta.get("skip_sync", False):
+        return None
```
ashley-hebler, line 877: **question:** If we're doing this in bulk, would it be at all helpful to log items we skip for debugging? Maybe we do that somewhere else so fine to ignore if so!

ashley-hebler (approved):
I added a few quick thoughts, but nbd on either since this is only used temporarily as you note

server/app.py
matthewdylan: yeah, I like that... fixed

server/app.py
matthewdylan: Yeah, very helpful to log these but I'm including that in the command that sends recurring donation records to become subscriptions instead.

PR #1145, opened 2023-12-11
add auto-rejected messaging in slack

matthewdylan:
#### What's this PR do?
If a donation that needs to be quarantined is auto-rejected, we show a message mentioning the auto-rejection. The usual info around the donation is included, but no further action will be needed.

Adding auto-reject logic has the ultimate goal of making life easier for the membership team by cutting down on the amount of quarantined donations.

Since there isn't a great way of mocking a bad recaptcha score, we'll just test that it doesn't block things it shouldn't be for now. To do that...

* visit the main donation page
* attempt giving a donation
* the donation should show as expected in #tech-test

We'll let membership know about the new feature.

Not here. I do wonder if we need a more informative message about the auto-rejection. It could be good to include more info as to why the rejection happened, but maybe that's unnecessary.

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/recelzGtohXuop8EK?blocks=hide

#### TODOs / next steps:

* [ ] deploy portal pr
* [ ] deploy this pr

### DIFF
--- a/server/bad_actor.py
+++ b/server/bad_actor.py
@@ -33,6 +33,7 @@ class BadActorResponseItem(BaseModel):
 
 class BadActorResponse(BaseModel):
     overall_judgment: BadActorJudgmentType
+    auto_rejected: bool
     items: List[BadActorResponseItem]
 
 
--- a/server/bad_actor.py
+++ b/server/bad_actor.py
@@ -117,7 +117,15 @@ class BadActor:
             },
         ]
 
-        if not self.bad_actor_api_response.auto_rejected:
+        if self.bad_actor_api_response.auto_rejected:
+            slack_block.append({
+                "type": "section",
+                "fields": [{
+                    "type": "mrkdwn",
+                    "text": "Donation auto-rejected"
+                }]
+            })
+        else:
             slack_block.append({
                 "type": "actions",
                 "block_id": "choices",
@@ -141,14 +149,6 @@ class BadActor:
                         "value": json.dumps(self.transaction_data),
                     },
                 ],
-            })
-        else:
-            slack_block.append({
-                "type": "section",
-                "fields": [{
-                    "type": "mrkdwn",
-                    "text": "Donation auto-rejected by CageAI"
-                }]
             })
 
         return slack_block
--- a/server/bad_actor.py
+++ b/server/bad_actor.py
@@ -106,7 +106,7 @@ class BadActor:
         info_items = self._slackify_items(info_items)
         judgment_items = self._slackify_items(judgment_items)
 
-        return [
+        slack_block = [
             {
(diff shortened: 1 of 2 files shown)


server/bad_actor.py
```diff
@@ -115,7 +115,10 @@ def _slackify_all(self) -> list:
                 "type": "section",
                 "fields": judgment_items,
             },
-            {
+        ]
+
+        if not self.bad_actor_api_response.auto_rejected:
```
ashley-hebler, line 120: **nitpick:** This is totally a matter of opinion, but this is kind of a double negative right? "If not not accepted..." For readability (at least in my brain!) it'd make more sense to swap these.

Also for whatever reason, I guess I'd expect for the "else" to be the more default behavior like:
if auto_reject:
     do something special
else:
     do the common thing
     
As will all things, there could be a good reason to have the if/else in this order that I'm not thinking of so open to your thoughts!

server/bad_actor.py
```diff
@@ -138,8 +141,17 @@ def _slackify_all(self) -> list:
...
                 ],
-            },
-        ]
+            })
+        else:
+            slack_block.append({
+                "type": "section",
+                "fields": [{
+                    "type": "mrkdwn",
+                    "text": "Donation auto-rejected by CageAI"
```
ashley-hebler, line 150: **note:** lol at CageAI. To your note in smells, that could make it sound a little more magical than it really is. I think leaving it as "Donation auto-rejected" would be just fine, although I do think this is a fun easter egg so also fine to leave it. I'll leave it to you! 🐣

ashley-hebler (approved):
Left a small nitpick note, but things LGTM on this end!

I really like your thought here:
> It could be good to include more info as to why the rejection happened, but maybe that's unnecessary.

That doesn't feel like a blocker, but you bring up a solid point about documenting this in some way. Maybe as a follow up to this we could update [this wiki doc](https://texastribune.atlassian.net/wiki/spaces/TECH/pages/907018241/Bad+Actors+card+testing) with this new business rule.

server/bad_actor.py
matthewdylan: No, I think that's a good point. In my head I still wanted the actions block to be handled first, but I do think this reads better and there's really no good reason to do it the other way.

server/bad_actor.py
matthewdylan: Yeah, that was more of a tongue-in-cheek commentary on the AI craze in general, but I see your point that that's not going to be the common understanding. Haha.

ashley-hebler:
I spotted some issues on the portal side, but this is still looking like it's in good shape! Thanks for the changes

PR #1147, opened 2023-12-12
Add rdo_converter command

matthewdylan:
#### What's this PR do?
Adds a click command that offers options for getting recurring donations from salesforce, paring them down to those not currently tracked by stripe and converting them to stripe subscriptions.

This is one of the last steps in our ongoing move of recurring donations to stripe subscriptions. We've been creating stripe subscriptions for months now (instead of recording them in salesforce and letting the nightly load pass charges to stripe when necessary). Converting these recurring donations to stripe subscriptions will allow them to be tracked and charged by stripe's system so that we aren't reliant on old technology on salesforce that we need to upgrade anyway (this helps in that upgrade process as well).

We're going to do a live demo/test of this post-standup so that anyone interested can look over my shoulder at the conversion process and make sure nothing seems amiss.

This change in particular doesn't need to be communicated, but we'll be in talks with the membership team when a conversion of a set of donation records happens.

Only that this command is temporary and that the --limit option is your friend, since you can do very small batch conversions to make sure a given conversion goes through as expected.

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/recJsJbQNe6QFgNMw?blocks=hide

#### TODOs / next steps:

* [ ] *your TODO here*


server/app.py
```diff
@@ -1603,3 +1610,79 @@ def close_rdo(subscription_id):
...
+
+            date = None
+            opps = rdo.opportunities()
+            for opp in opps:
+                if opp.stage_name == "Pledged":
+                    date = datetime.strptime(opp.close_date, "%Y-%m-%d")
+                    break
+            
+            if not date:
+                raise Exception("A proper date couldn't be found for RDO {rdo.id}, making it ineligble for conversion.")
```
djpeacher, line 1667: **question:** Does this mean this donation was either a one-time or a canceled recurring?

djpeacher (commented):
Looks good @matthewdylan, props for juggling all this 🤹🏼  I didn't run the code, but I traced it out, and the workflow overall makes sense to me. That tied with your demo makes me feel good about this.

I did leave a question or two, and the only thing that might be blocking is the `--only-staff` stuff.

server/app.py
```diff
@@ -1603,3 +1610,79 @@ def close_rdo(subscription_id):
...
+    # If any of the vars passed here are null, they're ignored and the query
+    # is built without them
+    rdos = RDO.list(limit=limit, donation_type=donation_type, email=email)
+    staff_names = []
+
+    # is there a better way to get a list of staff?
+    if only_staff:
+        response = requests.get('https://www.texastribune.org/api/v2/authors/?staff=true&limit=100')
+        staff_results = response.json()['results']
+        staff_names = [staff["name"] for staff in staff_results]
```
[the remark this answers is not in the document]
matthewdylan: Yeah, I'll chat with @ashley-hebler, but it might make sense to update this to grab any recurring donations with @texastribune.org emails affiliated as a next step. The only way we'd be able to go off of non-tribune related emails tied to staff is if we got a list of names from membership, and that might be more work all-around than it's worth.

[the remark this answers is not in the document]
ashley-hebler: Following along! It'd be worth asking if Kassie has that info easily accessible. In the end, if we don't find every staff member currently managed by salesforce, that's not the end of the world. We're going to move everyone eventually anyway.

server/util.py
```diff
@@ -242,3 +250,63 @@ def send_email_new_business_membership(account, contact):
...
     name_array = name.split(" ", 1) if name else ["", ""]
     return name_array[0], name_array[1]
+
+
+def amount_to_charge(form=None, amount=None, pay_fees=None, interval=None) -> Decimal:
+    """
+    Determine the amount to charge. This depends on whether the payer agreed
+    to pay fees or not. If they did then we add that to the amount charged.
+    Stripe charges 2.2% + $0.30 on single charges and 2.2% + 0.5% + $0.30
+    on recurring charges.
```
ashley-hebler, line 260: **question:** I remember way back that moving to stripe subscriptions meant an extra charge. Is that what the .5% is?

ashley-hebler (commented):
Makes sense at a high level. One bigger question is, will this need to drastically change once we start to do other subscription types? Since this is a weird code change we won't actually commit to master, would it make sense to merge this into a "stripe-migration" branch or something so you can tack on the dynamic parts that come with circle, blast, business?

server/util.py
matthewdylan: That's right. Now that you mention it, I think I'll need to add a way of telling donation_adder if the donation in question is new or legacy because our plan, up to now, was that legacy recurring donations should not be on the hook for this new .5%. Let me know if you think that's changed though @ashley-hebler

server/util.py
```diff
@@ -242,3 +250,63 @@ def send_email_new_business_membership(account, contact):
...
+        total = amount
+    return quantize(total)
+
+
+def quantize(amount):
+    return Decimal(amount).quantize(TWOPLACES)
+
+
+def donation_adder(customer: str, amount: int, pay_fees: bool, interval: str, year: int, month: int, day: int) -> object:
+    final_amount = amount_to_charge(amount=amount, pay_fees=pay_fees, interval=interval)
```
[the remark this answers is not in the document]
matthewdylan: amount is the initial amount given by the donor (pre-fees), final amount includes fees (in the case where they indicated they wanted to pay fees) and the int(final_amount * 100) bit is moving the amount to a non-decimal form (so 10.81 becomes 1081)... this is how stripe takes the amount... there's probably a name for that format that I could use to clarify that. I can clarify or notify amount vs. final_amount as well.

server/app.py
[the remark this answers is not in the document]
ashley-hebler: @matthewdylan lets just stick to tt.org to make our lives eaiser!

server/util.py
ashley-hebler: that's correct! We should not increase those fees for legacy recurring donations. I'll mention this to Liam as a reminder for our January conversion process

matthewdylan: Alright, I've added some logic to enforce this.

matthewdylan:
> Makes sense at a high level. One bigger question is, will this need to drastically change once we start to do other subscription types? Since this is a weird code change we won't actually commit to master, would it make sense to merge this into a "stripe-migration" branch or something so you can tack on the dynamic parts that come with circle, blast, business?

Forgot to address this yesterday, but yeah, I think that's a good idea and I've switched this pr to merge into a new branch instead of master now.

server/util.py
ashley-hebler: nice! I feel like that was the last missing piece in this so I'll go ahead and mark this as approved for me 🚀

ashley-hebler (approved):

PR #1162, opened 2024-01-25
correct link to printable membership form

matthewdylan:
#### What's this PR do?
Updates a link to point to a newer version of the printable membership form.

Irma reached out that a new member had sent in this outdated form and was wondering how they got it. Best guess is that they printed it or bookmarked the link back when it was prominently on our main donation page, but quickest way to dot our i's is to replace the old form link with the new one anywhere it exists in the code.

N/A

N/A

Can't figure out where this block of code would be showing up, at a glance, but switching to the newer form in the short term. Can look into if this is just unused code we can toss.

N/A (too small to really justify a ticket)

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/server/templates/includes/form_inner.html
+++ b/server/templates/includes/form_inner.html
@@ -33,5 +33,5 @@
     <button id="applePayButton" class="col_6 button button--l button--applePay">&nbsp;</button>
   </div>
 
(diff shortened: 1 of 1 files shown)


ashley-hebler (approved):
Yeah good call to update here for good measure. I also don't see where this is used so it's not clear how someone would find it, but I say :shipit:

PR #1170, opened 2024-02-09
Add more cancellation info

matthewdylan:
#### What's this PR do?
It fills in some info on recurring donation cancellations that the membership team have asked for, namely...
* adds a cancel date to the RDO object (no matter what the cause)
* adds a cancel date to the Contact object (when cancelled directly by a member)
* adds a cancel source (either "Member Portal" when triggered directly by member or "Staff" otherwise)
* messages the #bot-cancellations channel when a donation is cancelled

This will give more knowledge and awareness around recurring donation cancellation patterns and habits.

* On staging, add a new donation [here](https://donations-testing.herokuapp.com/donate) using the email you most often log into the portal with
* On staging, add another new donation using a different email that you don't often use to log into the portal

**Cancelling your own membership:**
* Log into the portal using the same email from step 1 and click memberships
* You should see the donation you made earlier listed, but if not, give it some time and refresh after a bit
* Cancel the donation you made earlier (no need to fill out the reason form)
* Bounce over to salesforce and find the recurring donation you just cancelled (quickest way is to search by the email you used and a table of recurring donations tied to that email should be the first thing you see)
* When you're on that record's page, you should see two fields...
    * "Cancellation Date" should have today's date
    * "Cancellation Method" should say "Member Portal"
* Over in slack land, a message should have populated on the #test-tech channel about the cancellation
 
**Cancelling another member's donation:**
* Still in the membership portal, use the tool that allows you to view as another member and enter the email you used in step 2 (above)
* From here largely follow the same steps for this donation as you did the last one with one caveat...
    * "Cancellation Method" should say "Staff"

We'll let the membership team know when it's done.

There's probably more discussion to be had around the date field on Contact... if a member cancels their donation and a week later gives a new recurring donation, do we need to wipe that field? How about if there are two rdos on one customer record (not likely but can happen)?

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/recDfgKWcBEkyfuZH?blocks=hide

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -1649,11 +1649,12 @@ def log_opportunity(contact, payment_intent):
 
 def close_rdo(subscription_id, user_initiated=False, contact=None):
     rdo = RDO.get(subscription_id=subscription_id)
-    rdo_update_details = {"npe03__Open_Ended_Status__c": "Closed", "Cancellation_Date__c": datetime.now()}
+    today = datetime.now(tz=ZONE).strftime("%Y-%m-%d")
+    rdo_update_details = {"npe03__Open_Ended_Status__c": "Closed", "Cancellation_Date__c": today}
     if user_initiated:
         rdo_update_details["Cancellation_Method__c"] = "Member Portal"
 
-        contact_update_details = {"Requested_Recurring_Cancellation__c": datetime.now()}
+        contact_update_details = {"Requested_Recurring_Cancellation__c": today}
         Contact.update([contact], contact_update_details)
 
     response = RDO.update([rdo], rdo_update_details)
--- a/server/app.py
+++ b/server/app.py
@@ -1653,7 +1653,7 @@ def close_rdo(subscription_id, user_initiated=False, contact=None):
     if user_initiated:
         rdo_update_details["Cancellation_Method__c"] = "Member Portal"
 
-        contact_update_details = {"Requested_Recurring_Cancellation__c": True}
+        contact_update_details = {"Requested_Recurring_Cancellation__c": datetime.now()}
         Contact.update([contact], contact_update_details)
 
     response = RDO.update([rdo], rdo_update_details)
--- a/server/app.py
+++ b/server/app.py
@@ -969,15 +969,19 @@ def payment_intent_succeeded(event):
 def customer_subscription_deleted(event):
     subscription = stripe.Subscription.retrieve(event["data"]["object"]["id"], expand=["customer"])
     customer = subscription["customer"]
-    details = subscription["cancellation_details"]
+    method = subscription["cancellation_details"].get('comment', 'Staff')
+    reason = subscription["cancellation_details"].get('reason')
 
     contact = Contact.get(email=customer["email"])
-    rdo = close_rdo(subscription["id"], user_initiated=details["comment"], contact=contact)
+    rdo = close_rdo(subscription["id"], method=method, contact=contact)
+    text = f"{contact.name}'s ${rdo.amount}/{rdo.installment_period} donation was cancelled due to {reason}"
+    if reason == "cancellation_requested":
+        text += f" ({method})"
 
     message = {
-        "text": f"{contact.name}'s ${rdo.amount}/{rdo.installment_period} donation was cancelled due to {details['reason']}",
+        "text": text,
         "channel": "#tech-test",
-        "icon_emoji": ":no-good:"
+        "icon_emoji": ":no_good:"
     }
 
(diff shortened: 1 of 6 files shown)


server/app.py
```diff
@@ -966,8 +967,24 @@ def payment_intent_succeeded(event):
...
+    subscription = stripe.Subscription.retrieve(event["data"]["object"]["id"], expand=["customer"])
+    customer = subscription["customer"]
+    method = subscription["cancellation_details"].get('comment', 'Staff')
+    reason = subscription["cancellation_details"].get('reason')
+
+    contact = Contact.get(email=customer["email"])
+    rdo = close_rdo(subscription["id"], method=method, contact=contact)
+    text = f"{contact.name}'s ${rdo.amount}/{rdo.installment_period} donation was cancelled due to {reason}"
+    if reason == "cancellation_requested":
+        text += f" ({method})"
```
ashley-hebler, line 979: **suggestion (non-blocking):** is reason always `cancellation_requested`?  If so, we could make this a little more succinct/scannable and have it read:

Cancellation: Ivana Cancel $35.00/monthly (Staff) 
-or-
Cancellation: Ivana Cancel $35.00/monthly (Member Portal)

server/app.py
```diff
@@ -966,8 +967,24 @@ def payment_intent_succeeded(event):
...
+
+    message = {
+        "text": text,
+        "channel": "#tech-test",
```
ashley-hebler, line 983: **polish:** this would be cool to have as a config var. Maybe you were already scheming that

server/app.py
```diff
@@ -966,8 +967,24 @@ def payment_intent_succeeded(event):
...
+        "icon_emoji": ":no_good:"
```
ashley-hebler, line 984: **praise:** haha excellent emoji curation here

ashley-hebler (approved):
Test steps all WFM! I left some notes, but nothing seems particularly blocking.

From your PR note here...
> There's probably more discussion to be had around the date field on Contact... if a member cancels their donation and a week later gives a new recurring donation, do we need to wipe that field? How about if there are two rdos on one customer record (not likely but can happen)?

... I feel like when I donated as the same account, they all got counted as separate RDOs and then as I cancelled each, through portal, staff, and stripe, they were all handled separately with the right values. That seems like exactly what we want in my head, but I could also be testing for the wrong scenario.

Speaking of which, cancelling through Stripe is one where I noticed nothing came through for the "method." Doesn't seem like the end of the world, but if it's easy enough to add some better label there that might be nice.

All in all, the observability here is a great improvement!

server/app.py
matthewdylan: No, and that's a good question that leads to some further info. Though not something I can test per se, this update should also end up pushing cancellations due to card failures and the like to slack. I'm assuming that would be good to see in that channel, but let me know if you think it's overkill and I can re-write this to kick those out. I figured it would be better to start with sending any cancellations (whether user/staff initiated or from charging failure) to that channel and if it's too much noise, we can reduce, but I'm interested in your opinion.

server/app.py
matthewdylan: Oh yeah, good point. Wouldn't be very helpful going there in prod :facepalm: Can fix.

server/app.py
matthewdylan: 🙇

matthewdylan:
> Test steps all WFM! I left some notes, but nothing seems particularly blocking.
> 
> From your PR note here...
> 
> > There's probably more discussion to be had around the date field on Contact... if a member cancels their donation and a week later gives a new recurring donation, do we need to wipe that field? How about if there are two rdos on one customer record (not likely but can happen)?
> 
> ... I feel like when I donated as the same account, they all got counted as separate RDOs and then as I cancelled each, through portal, staff, and stripe, they were all handled separately with the right values. That seems like exactly what we want in my head, but I could also be testing for the wrong scenario.
> 
> Speaking of which, cancelling through Stripe is one where I noticed nothing came through for the "method." Doesn't seem like the end of the world, but if it's easy enough to add some better label there that might be nice.
> 
> All in all, the observability here is a great improvement!

Oh, that great news. Thanks for the review!

server/app.py
ashley-hebler: We talked about this in realtime, but I think this is sound reasoning 🤝

ashley-hebler (approved):
New changes WFM!

PR #1174, opened 2024-02-21
update closed lost reason on next opp

matthewdylan:
#### What's this PR do?
* adds an order_by to the opportunities func on RDO for ease of grabbing the next pledged opp
* updates the stage and closed lost reason on the next pledged opp
* adds better error handling and messaging

The salesforce side is expecting a closed lost reason for cancelled subscriptions due to payment failures.

* Start a [new stripe clocks simulation](https://dashboard.stripe.com/test/billing/subscriptions/test-clocks) (call it whatever you like and go with the current date)
* Screen Shot 2024-02-26 at 10 18 17 AM
* Once you're looking at your simulation, add a new user (add dropdown menu halfway down the page)
* Screen Shot 2024-02-26 at 10 21 28 AM
    * include name, email and currency (USD)
    * make sure the email is one you've previously used on donations staging (you can use "matthew.mccrary@texastribune.org" or "ashleyhebler+2-22-24@gmail.com" if you're just sitting there staring at this line)
    * Screen Shot 2024-02-26 at 10 21 48 AM
        * context -> stripe clocks doesn't give enough information to the salesforce side to create a brand new contact, though it does give enough info to create a brand new subscription... ergo, using an email that is already tied to a contact in our salesforce sandbox allows a new rdo to be added to an existing contact (otherwise this step fails)
* Now click on the newly created customer and add a new subscription
* Screen Shot 2024-02-26 at 10 39 15 AM
    * Keep the starting date as is (current date by clock)
    * For "Pricing", search for "blast" and pick one of the products listed
    * Screen Shot 2024-02-26 at 10 24 53 AM
    * IMPORTANT: in the "Subscription Settings - Payment" area, add a new payment method (card) using any of the [stripe test cards](https://stripe.com/docs/testing#cards)
        * if you see the option to set the card as Default, do so
        * Screen Shot 2024-02-22 at 12 13 03 PM
    * Choose "Schedule subscription" at bottom of the form
* You should see the new scheduled subscription listed under "Subscriptions"

* Since we've had issues with stripe clocks to this point, I would check [Salesforce](https://texastribune--2023stripe.sandbox.lightning.force.com/lightning/page/home) to make sure the subscription synced over there as an RDO (doing a search by the email you used or the stripe customer_id is probably the quickest way to find the RDO, and the RDO title will include the time, so go off of that to confirm you're looking at the right one)
* Screen Shot 2024-02-26 at 10 31 01 AM
* Found it? Take a breath. You're through the most annoying phase of this (I think 😅)

* Back in stripe, find the subscription you just created and click on "Update Subscription" from the drop-down menu
* Screen Shot 2024-02-22 at 12 23 18 PM
* Find the "Subscription Settings - Payment" section again and click to edit the payment
* Choose to add a new card and use the exact card number in the following screenshot (this one in particular is meant to pass stripe's initial check during payment update but purposefully fails at time of charge) and make sure "Set as customer's new default" is checked
* Screen Shot 2024-02-22 at 12 27 21 PM
* After saving, you can refresh the page to make sure the new card is set as "Default" in the "Payments Section" of the customer page
* Screen Shot 2024-02-22 at 12 27 54 PM
* Now, click on "Advance time" in the simulation toolbar at the top of the page
* Set the clock to a month and three+ hours in advance of whatever the clock currently says, then click "Advance"
* After advancing to this time, you should see the subscription is in a "Past due" status and the most recent Payment is marked "Failed". This is basically the first time the charge was attempted and failed.
* Because the charge is retried two more times over a span of nine days, now advance two weeks ahead (you could instead advance by nine+ days, but advancing two weeks ahead is just easier to see on the calendar)
* This should trigger the customer.subscription.deleted event which should both close the recurring donation AND update the next opportunity to have a Closed Lost Reason
    * From the recurring donation, click "Related" to get a list of the opportunities and click to see all opps (besides the very last opp you see, they should all be marked "Closed Lost" at this point)
    * Screen Shot 2024-02-26 at 10 52 56 AM
    * Click on the one closest from the bottom marked "Closed Lost"
    * Screen Shot 2024-02-26 at 10 53 19 AM
    * Now, looking at the opportunity, click "Details" and it's easiest to do a ctrl-f for "Closed Lost Reason" from here, because this page has a lot going on
    * Screen Shot 2024-02-26 at 10 53 40 AM
    * The "Closed Lost Reason" field should be filled (likely "payment_failed")
    * Screen Shot 2024-02-26 at 10 54 05 AM

We'll update the #project-donations channel where this was initially requested.

Not really. Some of this will code will be made moot when we migrate to the newer version of salesforce because an rdo will ONLY point to the next pledged opp, but most of it will still be applicable. Also, if it proves fruitful, we could shore up a more human friendly message about why the charge failed, though that would probably be more helpful in slack than in salesforce. (FWIW this info stays available in stripe, so we're not losing it by not syncing it t SF)

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/recHELo7XtiYHnJqA?blocks=hide

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -1580,9 +1580,11 @@ def log_rdo(type=None, contact=None, account=None, subscription=None):
     if source:
         card = stripe.Customer.retrieve_source(customer_id, source)
     else:
-        customer = stripe.Customer.retrieve(customer_id)
-        card = customer.sources.retrieve(customer.sources.data[0].id)
-        app.logger.info(f"in here too! {card}")
+        customer = stripe.Customer.retrieve(customer_id, expand=["sources"])
+        app.logger.info(customer)
+        card = customer["sources"]["data"][0]
+        # card = customer.sources.retrieve(customer.sources.data[0].id)
+        # app.logger.info(f"in here too! {card}")
 
     year = card["exp_year"]
     month = card["exp_month"]
--- a/server/app.py
+++ b/server/app.py
@@ -1581,6 +1581,7 @@ def log_rdo(type=None, contact=None, account=None, subscription=None):
         card = stripe.Customer.retrieve_source(customer_id, source)
     else:
         customer = stripe.Customer.retrieve(customer_id)
+        app.logger.info(f"in here! {customer.sources.data}")
         card = customer.sources.retrieve(customer.sources.data[0].id)
 
     year = card["exp_year"]
--- a/server/app.py
+++ b/server/app.py
@@ -923,7 +923,7 @@ def customer_subscription_created(event):
     # use that, otherwise retrieve the latest invoice from stripe
     if not subscription["trial_end"] and invoice_status != "draft":
         update_next_opportunity(
-            opps=rdo.opportunities(),
+            opps=rdo.opportunities(ordered_pledges=True),
             invoice=invoice,
         )
 
--- a/server/app.py
+++ b/server/app.py
@@ -904,7 +904,7 @@ def customer_subscription_created(event):
             contact.save()
 
         account = get_account(customer)
-        rdo = log_rdo(type=donation_type, account=account, subscription=subscription)
+        rdo = log_rdo(type=donation_type, account=account, customer=customer, subscription=subscription)
         affiliation = Affiliation.get_or_create(
             account=account, contact=contact, role="Business Member Donor"
         )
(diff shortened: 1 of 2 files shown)


djpeacher (approved):
Wow 😵 Those clock simulations are pretty cool (and easy to get turned around in). I only glanced at the code, but I went through the testing steps twice (once with a monthly, and once with a yearly) and both times I was able to produce the results you outlined (h/t btw for the steps/screenshots!).

PR #1182, opened 2024-05-03
fix stripe missing card issue during sync

matthewdylan:
#### What's this PR do?
Stops assuming that stripe properly passes back a card at time of RDO creation.

With the current way this is set up, instances where stripe DOESN'T properly send back the card info result in the rdo not being created at all. Longterm, we need to figure out why this data doesn't reliably come through, but in the meantime, rdo should always get created and we can throw up an error message that lets us know we need to update the card info manually.

Huh... good question. Since I can't reliably get stripe to not send card info, I don't really have a good way of testing this, besides getting it into production and rerunning a recent subscription that didn't sync properly when getting to this step because the card data can't be found.

We can notify Morgan about this fix.

This missing card thing seems to keep happening, which is weird because the charges are going through, so the card is there... I need to play around with stripe some (maybe even more fun with CLOCKS) to see why this info is missing sometimes (likely the timing is my guess) so there should be a future pr that has a better source for this data BUT, either way, good to make this a get instead of always assuming the data will be there.

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -1580,15 +1580,19 @@ def log_rdo(type=None, contact=None, account=None, customer=None, subscription=N
     else:
         card = stripe.Customer.retrieve_source(customer_id, customer["invoice_settings"]["default_payment_method"])
 
-    year = card["exp_year"]
-    month = card["exp_month"]
-    day = calendar.monthrange(year, month)[1]
+    year = card.get("exp_year", None)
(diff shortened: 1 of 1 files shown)


ashley-hebler (approved):
LGTM! Seems totally reasonable as a good solve for now

PR #1188, opened 2024-05-16
update log_opportunity for missing card

matthewdylan:
#### What's this PR do?
Basically a carbon copy of [this pr](https://github.com/texastribune/donations/pull/1182) except for the log_opportunity function (ref'd pr was for the log_rdo function).

This will keep the sync from stripe to SF from breaking just because stripe isn't looking in the right places for card data (specifically for opportunities)

It really can't be. This is a hard issue to recreate manually, but I'll re-run one of the recently blocked events after this is deployed to check that it worked.

It doesn't need to be.

We'll need to figure out why stripe isn't finding the card data correctly.

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/recg4FHGZnRzZP5US?blocks=hide

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -1649,16 +1649,21 @@ def log_opportunity(contact, payment_intent):
     opportunity.quarantined = True if payment_meta.get("quarantined", None) else False
 
     card = stripe.Customer.retrieve_source(customer_id, payment_intent["source"])
-    year = card["exp_year"]
-    month = card["exp_month"]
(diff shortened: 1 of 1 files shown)


ashley-hebler (approved):
Didn't test it obviously, but step through line by line and I'm not spotting any issues. LGTM! And good plan to merge then quickly test to see

PR #1190, opened 2024-05-23
fix unsubscriptable opportunity bug

matthewdylan:
#### What's this PR do?
treats the opportunity object like an object instead of a dict

the other way was breaking

Another hard one to test. I can rerun the push from stripe for a particular event that broke due to this bug and make sure it gets through the syncing flow as expected.

N/A

N/A

N/A


ashley-hebler (approved):
:shipit:

PR #1191, opened 2024-05-28
Add event syncer

matthewdylan:
#### What's this PR do?
Adds a callable function that takes an event id, grabs the event object from stripe and passes it through the usual stripe -> SF sync. Also added the ability to pass in a subscription id or payment intent id to the webhook function that handle those, in the case where the event mentioned above is out of retention (stripe only has a 30 day retention policy for events for some reason).

In cases where events crashed out of the stripe->SF sync flow, we can use these tools to still push that info to SF, keeping every donation logged.

Test out a few regular gifts
1) make sure they show up on slack
2) make sure they show up on salesforce

After this, turn off the webhook by...
1) In the stripe dashboard, make sure you're in test mode by toggling it in the upper right-hand corner
    a) extremely important... if you do the next steps and you're NOT in test mode, the sync from stripe to salesforce will effectively be broken (at least temporarily)
2) Click on the "Developers" button just to the left of the test mode toggle
3) Click on the "Webhooks" tab
4) Click on the url listed under "Hosted Endpoints"
5) Click on the menu button on the right-hand side of the page (an ellipses) and choose "Disable..."
    a) This breaks the connection between donations given through staging and the sf db for staging
6) Make another donation
7) Find the donation on stripe by going to the Payments section (still in test mode) and clicking the first entry
8) Copy the customer id on the right of the page and search for it in salesforce
    a) You shouldn't be able to find anything
9) Copy the event id (in the url)
9) Visit our heroku setup for donations-testing
10) On the right hand side of the dashboard click the "More" dropdown and choose "Run console"
    * Screen Shot 2024-05-28 at 12 10 43 PM
11) When the heroku console is pulled up, run "flask --app server shell" which drops you into the flask shell
12) Run "from server.app import sync_stripe_event"
13) Run "sync_stripe_event(event_id)" (using the event id you copied earlier)
14) Check back on the salesforce side, search for the customer id again and now you should see the recurring donation listed

We can let Morgan know that we're ready to fix any missing donations.

One thing to note is that it would be nice to be able to call the new function from server.util instead of server.app, but that would mean bringing several other functions with it to util.py (pretty much all of the new functions written last year in the transaction to stripe subscriptions). I started toying around with this, but then it ended up being over-complicated for the task at hand so I pulled back and kept this one more simple.

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/recUzjSnuXgxRUTug?blocks=hide

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -1549,19 +1549,19 @@ def log_rdo(type=None, contact=None, account=None, customer=None, subscription=N
     sub_meta = subscription["metadata"]
     sub_plan = subscription["plan"]
     customer_id = subscription["customer"]
-    trial_end = subscription["trial_end"]
+    start_date = subscription["start_date"]
     donation_type_info = DONATION_TYPE_INFO[type]
     installment_period = "yearly" if sub_plan["interval"] == "year" else "monthly"
 
     if type == "business_membership" and account:
-        rdo = RDO(account=account)
+        rdo = RDO(account=account, date=start_date)
         year = datetime.now(tz=ZONE).strftime("%Y")
         rdo.name = f"{year} Business {account.name} Recurring"
         rdo.record_type_name = "Business Membership"
         rdo.installments = None
 
     else:
-        rdo = RDO(contact=contact, date=trial_end)
+        rdo = RDO(contact=contact, date=start_date)
         rdo.installments = None
         if type == "circle":
             rdo.installments = 36 if sub_plan["interval"] == "month" else 3
--- a/server/app.py
+++ b/server/app.py
@@ -888,7 +888,7 @@ def customer_subscription_created(subscription_id):
 
     invoice_status = invoice["status"]
     if invoice_status == "open":
-        raise Exception(f"Subscription {subscription["id"]} was created but its first invoice is still open.\
+        raise Exception(f"Subscription {subscription['id']} was created but its first invoice is still open.\
                         Please follow up with the subscription to proceed.")
     customer = stripe.Customer.retrieve(subscription["customer"])
     contact = get_contact(customer)
--- a/server/app.py
+++ b/server/app.py
@@ -1146,6 +1146,12 @@ def stripehook():
 
     app.logger.info(f"Received event: id={event.id}, type={event.type}")
 
+    process_stripe_event(event)
+
+    return "Success", 200
+
+
+def process_stripe_event(event):
     # setting celery's delay on these keeps the stripe call from erroring out
     if event.type == "customer.source.updated":
         customer_source_updated.delay(event)
@@ -1161,7 +1167,15 @@ def stripehook():
     if event.type == "subscription_schedule.updated":
         subscription_schedule_updated.delay(event)
 
-    return "Success", 200
+    return True
+
+
+def sync_stripe_event(event):
+    stripe_event = stripe.Event.retrieve(event)
+
+    process_stripe_event(stripe_event)
+
+    return stripe_event
 
 
 # this is just a temp func version of a piece of add_donation we're
--- a/server/app.py
+++ b/server/app.py
@@ -867,10 +867,10 @@ def payout_paid(event):
 
 
 @celery.task(name="app.customer_subscription_created")
-def customer_subscription_created(event):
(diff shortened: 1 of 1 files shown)


server/app.py
```diff
@@ -1146,22 +1138,39 @@ def stripehook():
...
+
+def process_stripe_event(event):
     # setting celery's delay on these keeps the stripe call from erroring out
     if event.type == "customer.source.updated":
         customer_source_updated.delay(event)
     if event.type == "payout.paid":
         payout_paid.delay(event)
     if event.type == "customer.subscription.created":
-        customer_subscription_created.delay(event)
+        subscription_id = event["data"]["object"]["id"]
```
ashley-hebler, line 1153: **suggestion: (non-blocking)** Could we store these as variables higher up? Or is it that sometimes an event won't have these keys in the event?

Something like:
```py
event_data = event["data"]["object"]
event_type = event.type # while we're at it
...
if event_type == "customer.subscription.created":
     subscription_id = event_data["id"]
```

server/app.py
```diff
@@ -933,23 +932,17 @@ def customer_subscription_created(event):
...
-    payment_intent = event['data']['object']
-    invoice_id = payment_intent['invoice']
-    if invoice_id:
-        try:
-            invoice = stripe.Invoice.retrieve(invoice_id, expand=['subscription'])
-        except stripe.error.StripeError as e:
-            app.logger.error(f"Issue finding invoice for {payment_intent['id']} with message: {e}")
-            return # process can not continue without invoice and will need to be retriggered from stripe
+def payment_intent_succeeded(payment_intent_id):
+    payment_intent = stripe.PaymentIntent.retrieve(payment_intent_id, expand=["invoice.subscription"])
```
ashley-hebler, line 936: **suggestion (non-blocking):** Might be overkill, but would it make sense to wrap this in a try/except, similar to how you handled the invoice retrieval previously?

server/app.py
```diff
@@ -1535,19 +1544,19 @@ def log_rdo(type=None, contact=None, account=None, customer=None, subscription=N
...
-        rdo = RDO(account=account)
+        rdo = RDO(account=account, date=start_date)
         year = datetime.now(tz=ZONE).strftime("%Y")
         rdo.name = f"{year} Business {account.name} Recurring"
         rdo.record_type_name = "Business Membership"
         rdo.installments = None
 
     else:
-        rdo = RDO(contact=contact, date=trial_end)
+        rdo = RDO(contact=contact, date=start_date)
```
ashley-hebler, line 1559: **question:** Just more curious than anything, but what does this do?

ashley-hebler (approved):
All test steps WFM! This is a great utility tool to have even if it is a little high touch. Makes me appreciate that django builds in management command functionality!
It does feel like a process we should document outside of this PR. Maybe our wiki could be a good spot.

I left a few take-it-or-leave it notes and one question, but nothing looked blocking to me. Thanks for doing this! Our future selves (and the accounting team!) will be grateful.

server/app.py
matthewdylan: Great catch. Fixed.

server/app.py
matthewdylan: I don't think it's necessary. The reason why I had that in there previously is because I did run into situations where a payment_intent event object would have an invoice_id but the invoice wouldn't be available from stripe for whatever reason. I think this was simply a timing thing, whereas here we're calling with a payment_intent that we know has already been processed since it was the thing that was passed to us. There's probably a better way of explaining that, so we can talk about it in our 1:1 if you want to.

server/app.py
matthewdylan: This code is passing the start_date to the RDO so that the RDO can better reflect the subscription it's dealing with. This is important in cases where we're syncing an old subscription that never made it to Salesforce so that the recurring dates in SF line up with the recurring dates in Stripe. Currently, the RDO just takes today's date (or the trial_end as you see there, but that's no longer relevant).

server/app.py
ashley-hebler: All good, WFM!

PR #1192, opened 2024-06-20
"Reason" query param

ashley-hebler:
#### What's this PR do?

Adds the value of "reason" to the form if passed in a query param

Screenshot 2024-06-20 at 4 30 56 PM

This is useful for special donations that need to be classified consistently

Confirm `http://local.texastribune.org/donate?reason=foo` inputs `foo` into the "I'm giving because" field

I'll let the development team know

I don't think so. This was a backlogged item from last year and when I re-looked into it I realized it was comically easy. Made me wonder if I was missing something

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viw2eOTqOWglhM7UI/recQdSKsn2hErH1YG

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/server/static/js/src/entry/donate/router.js
+++ b/server/static/js/src/entry/donate/router.js
@@ -32,6 +32,7 @@ function createInitialFormState(queryParams) {
     email = '',
     zipcode = '',
     code = '',
+    reason = '',
   } = cleanParams;
   let amount;
   let installmentPeriod;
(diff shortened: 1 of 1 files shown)


matthewdylan (approved):
LGTM! Easy win.

PR #1202, opened 2024-07-30
adding default campaign to Waco

matthewdylan:
#### What's this PR do?
adds a default campaign id for any donations coming through the /waco landing page

Helps us to keep track of these donations and to group them with the donations we've already gotten for Waco tied to this same campaign.

Attempt giving a recurring donation on staging
Find the subscription in stripe and look at the metadata
You should see a campaign_id key with a value

We'll let Waco know when the email for this campaign is for sure working.

Not my favorite way of handling this, but it'll work in the short term.

#### TODOs / next steps:

* [ ] *your TODO here*


ashley-hebler (approved):
WFM!

PR #1203, opened 2024-08-06
less strict syncing

matthewdylan:
#### What's this PR do?
* makes it so we can get around a strict exception if needed
* also makes it so the MAX_SYNC_DAYS_DIFFERENCE var can be controlled from heroku (probably should have always been this way)

The exception in question is good to know about, but once we figure out what's going on, we want to be able to rerun a donation through the salesforce sync and expect that the sync gets around the exception.

Sadly, this is another one of those edge cases that would be overly difficult to replicate in a testing scenario. Suffice it to say, this update won't effect the regular donation/sync flow and I'll rerun the donation in question after this is in place to make sure it works as expected.

Doesn't need to be.

Naw.

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/recmm6dVQibvmzmO1?blocks=hide

### DIFF
--- a/app.json
+++ b/app.json
@@ -105,6 +105,9 @@
     "MAILTRAP_API_TOKEN": {
       "required": true
     },
+    "MAX_SYNC_DAYS_DIFFERENCE": {
+      "required": false
+    },
     "MULTIPLE_ACCOUNT_WARNING_MAIL_RECIPIENT": {
       "required": true
     },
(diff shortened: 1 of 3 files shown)


ashley-hebler (approved):
Stepped through the code and LGTM! Good call to make MAX_DAYS... a config var.

PR #1220, opened 2024-10-28
Add config elements for redis upgrade

matthewdylan:
#### What's this PR do?
Fixes an error encountered with the redis upgrades prompted by heroku.

As is, once the redis upgrade is pushedd by heroku in production, our site will break without these config elements. The staging site was broken before this branch was pushed there.

// These first steps make sure local development isn't effected
For now, make sure you have the full sandbox salesforce login info setup in env-docker
Run make locally
Ensure that a donation can be given

// These next steps make sure staging is working
Visit [the staging site](https://donations-testing.herokuapp.com/donate)
Ensure a donation can be given

It doesn't need to be

I went back and forth on the different CERT_ options and did some research on what we could do here. Ultimately, I went with the option that is most similar to what we currently have and is the simplest fix for the site being non-functional. That said, the CERT_ option here should be seen as a temporary one until further work can be given to our heroku redis ssl setup. These updates don't make things less secure, but with this new celery setting, we have the ability to make things more secure moving forward, given some time and research.

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/recqp0mdwTmngDIMJ?blocks=hide

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -13,3 +13,4 @@ services:
     entrypoint: python -m pytest "tests" --cov=.
     environment:
       - SECRET_KEY=sk_test_ok
+      - REDIS_URL="http://redis.test.com"
--- a/server/app_celery.py
+++ b/server/app_celery.py
@@ -11,7 +11,7 @@ def make_celery(app):
         broker=app.config["CELERY_BROKER_URL"],
         broker_connection_timeout=60,
         redis_backend_use_ssl = {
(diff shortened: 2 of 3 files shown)


ashley-hebler (approved):
Staging testing WFM!

I'd be fine with expediting this deploy if we do find that it's impacting the ability to give in Waco

PR #1223, opened 2024-11-01
Fix REDIS_URL in tests

ashley-hebler:
#### What's this PR do?

Adds an env var `REDIS_URL` to the circle build

We introduced a breaking commit in #1220 

N/A just let circle do its thing

N/A

@matthewdylan I remember that environment bit in the docker-compose.test.yml because it seemed like that wasn't doing anything. Let me know if I missed a detail there though.

off-sprint, but wanted to get this working for upgrading the heroku stack


matthewdylan (approved):
Nothing to see here, really. Tests pass finally (THANK YOU @ashley-hebler!) and working locally. GG!

PR #1230, opened 2024-11-27
Remove a/b testing check for dynamic donation amounts

matthewdylan:
#### What's this PR do?
Removes the check for an a/b testing query so that these dynamics amounts show for anyone on the page.

Makes this feature work for anyone visiting the donate page.

Visit /donate
Ensure the amount changes when switching between monthly, yearly and one-time
Ensure the amount DOES NOT change after manually changing the amount

Doesn't need to be.

Not for this in particular.


ashley-hebler (approved):
Works as intended!

PR #1234, opened 2024-12-10
Add new blast page

matthewdylan:
#### What's this PR do?
Adds a new page/form for Blast subscriptions.

We wanted to add some new options for the Blast and creating this page at the same time brings the Blast workflow closer in accordance with the rest of the donations pages.

Visit the blast page on [staging](https://donations-testing.herokuapp.com/blast)
For each option (standard, tax-exempt, academic and legislative), for monthly, yearly or session as applicable...
    * create a new donation
    * check for that donation in stripe (customers/transactions tabs)
        * check that the amount and cadence on the item is accurate
    * check that the donation has synced through to salesforce (can look up by stripe subscription or customer id)
    * on the recurring donation object in salesforce, ensure the subscriber email is listed

I'm assuming this part falls to @ashley-hebler and Ryan Kim, but happy to help as needed.

A few things...
* I spent some time (probably more than I should have) trying to get some stripe elements working here to make the front-end's access to products/prices easier, cleaner and more efficient, but without much luck. When I to have more time, I'd really like to get that working
* The new legislative session-only option is set on a two-year cadence in stripe, which means that the user will only get charged every session (since that's on a two-year cadence as well). There are definitely other ways we could do this (it could just be a one-time charge, if that works) but I felt like this was more in line with the spirit of the option on the front-end (240 per session). The only hang-up here is that, to my knowledge, this same cadence isn't available in SF, so we end up on a yearly cadence there, meaning we would need to manually close those opportunities that pop-up in the off year (2026, 2028). I'll reach out to Anna to see if I missed a way of doing this more properly in SF.

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/recbnhUrrBplaw7Gn?blocks=hide

#### TODOs / next steps:

* [ ] *your TODO here*

### DIFF
--- a/server/static/js/src/entry/blast/constants.js
+++ b/server/static/js/src/entry/blast/constants.js
@@ -72,7 +72,7 @@ export const BLAST_FORM_STATE = {
     message: null,
   },
   description: {
-    value: 'The Texas Tribune Circle Membership',
+    value: 'The Blast Subscription',
     isValid: true,
     validator: null,
     message: null,
--- a/server/static/js/src/entry/blast/constants.js
+++ b/server/static/js/src/entry/blast/constants.js
@@ -38,11 +38,11 @@ export const BLAST_LEVELS = {
   blastLegislativeSession: {
     bucket: 'legislative',
     bucketFull: 'Legislative session-only',
-    installmentPeriod: 'session',
+    installmentPeriod: 'one-time',
     amount: '200',
     isFeatured: true,
     prompt: 'Get ahead with insider news!',
-    footer: 'January 17 – June 17',
+    footer: 'January 14 – June 14',
   },
 };
 
--- a/server/static/js/src/entry/blast/constants.js
+++ b/server/static/js/src/entry/blast/constants.js
@@ -1,0 +1,97 @@
+import * as validators from '../../utils/validators';
+
+export const DEFAULT_LEVEL = 'blastMonthly';
+
+export const BLAST_LEVELS = {
+  blastMonthly: {
+    bucket: 'standard',
+    bucketFull: 'Standard subscription',
+    installmentPeriod: 'monthly',
+    amount: '40',
+  },
+  blastYearly: {
+    bucket: 'standard',
+    bucketFull: 'Standard subscription',
+    installmentPeriod: 'yearly',
+    amount: '349',
+  },
+  blastTaxExemptYearly: {
+    bucket: 'tax-exempt',
+    bucketFull: 'Tax-exempt subscription',
+    installmentPeriod: 'yearly',
+    amount: '325',
+  },
+  blastAcademicMonthly: {
+    bucket: 'academic',
+    bucketFull: 'Academic discount',
+    installmentPeriod: 'monthly',
+    amount: '20',
+  },
+  blastAcademicYearly: {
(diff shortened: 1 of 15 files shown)


server/static/js/src/entry/blast/constants.js
```diff
@@ -0,0 +1,85 @@
...
+    message: 'Enter your last name',
+  },
+  pay_fees_value: {
+    value: 'False',
+    isValid: true,
+    validator: null,
+    message: null,
+  },
+  description: {
+    value: 'The Texas Tribune Circle Membership',
```
ashley-hebler, line 70: **question:** Does this appear anywhere?

ashley-hebler (commented):
This is spectacular! It's a relief to get the blast into the right form setup.

I have a few one-off suggestions and thoughts:

**Before launch:**
1. Could we update the form submit message? I wouldn't say it has to be a replica of the current blast submission page, but something with Blast branding and contact info.
2. I like that we have the old legacy for for safe keeping but I'd be in favor of some redirects from /blastform to /blast. That way we don't have to find all the places we're linking to to. We could also redirect /blast-promo to /blast for good measure

**To discuss:**
1. That's a really clever way to handle the legislative session option! Maybe even too clever in that people might forget about this charge. I'd be in favor of a one-time charge here just because that's what I would want as a subscriber and think Ryan and team said that'd be fine for this. That could be a stripe sub that just doesn't have a future payment or a just a plain ol' one time charge like the donate form. I defer to you and Anna on that!


**Design updates:**
1. Add value propositions
     - Add the strikethrough text for the ~$480~ $349 yearly (if possible! not a dealbreaker)
     - Add a yellow box below the standard sub with "Save 27% with yearly" text in it
     - Add some text above the lege option that says "Get ahead with insider news!"
          - You might have some luck with `position: absolute; top: -1rem` or something for this so that it doesn't mess with the vertical alignment
     - Add a #a1d2df border around the lege option
     - Add a #a1d2df box below the lege option with "January 17 – June 17" text in it
3. Remove academic discount for now (i think it'd be good to leave the backend stuff here for now, but just remove the front-end display of it)
4. Change the session only price to $200

^ if you want to pair on any of these let me know!

Here's a mockup of all of the above:
Quick Blast Mockup - Design revision

server/static/js/src/entry/blast/constants.js
matthewdylan: Oops, that's a leftover. I'll remove.

matthewdylan:
> @matthewdylan, one additional testing note: I noticed that when I tested in the old form, it came through to stripe without a subscription and possibly this [error](https://the-texas-tribune.sentry.io/issues/6126533703/?environment=test&project=1304832&query=is:unresolved%20issue.priority:%5Bhigh,%20medium%5D&statsPeriod=24h&stream_index=0): Screenshot 2024-12-10 at 1 13 21 PM
> 
> That ends up not mattering if we go ahead and redirect those old pages to the shiny new one. But just wanted to point that out!

Good note. I'll definitely redirect. That's an easy fix.

ashley-hebler (approved):
Did some testing on Friday and this LGTM!

PR #1239, opened 2025-01-28
Redirect Broken URL

djpeacher:
#### What's this PR do?
- Adds membership query params to error page.
- Redirects `/donate/` to `/donate`.

- Cleanup

- Confirm going to https://support.texastribune.org/donate/ redirects.

- Follow up with Emily.

- Query params are lost during the redirect.

- Off sprint


ashley-hebler (approved):
Phew much better. Thank you!

PR #1242, opened 2025-02-19
turn off dynamic amount

matthewdylan:
#### What's this PR do?
Changing frequency of donations will no longer trigger an amount change. 

We'll keep this turned off until we have a better way to display the final amount at time of charge. This will hopefully curb the issues we are seeing with donors accidentally giving more than they intend to.

* Spin up locally
* Hit /donate
* Changing between "Monthly donation", "Yearly donation" and "One-time donation" should have no effect on the amount

I'll let Morgan and Emily know

We want to eventually turn this off when we have better displaying of the final amount.


ashley-hebler (approved):
:shipit:

PR #1243, opened 2025-02-24
Add Waco business membership form

matthewdylan:
Gives a landing page for business membership signups.

* Can run locally but probably easiest to start on Waco staging's [business page](https://waco-donations-testing-1119390b0075.herokuapp.com/business)
* Try giving at different levels and track donations through the stripe -> salesforce -> slack flow

We'll let April and Melissa know since this request started with them.

There is plenty of room to clean this up. They needed a quick turnaround, so this was literally a plug-and-play to get the business page from TT to work on Waco, followed by a very rough clean-up.

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/rec3fbfROpJFkMJTf?blocks=hide

#### TODOs / next steps:

* [ ] *your TODO here*


ashley-hebler (approved):
All LGTM! It's kinda cool to see the Waco branding scaling this way. I support the approach of duplication for the sake of getting this out the door. Also if we're following the [rule of 3](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)), this is technically in the clear 😈 

Two flags:
- not code-related: Could you ping Anna C. about this if you haven't already?
- future change: Our thank you pages aren't really "product" aware, which has always been the case regardless of this PR. BUT I did notice that the waco one in particular refers to the "Waco Local News Initiative" which I think was just the name before the name. Fine to update that in a future PR IMO!

djpeacher (approved):
LGTM. Records appeared in Stripe, Salesforce, Slack as expected.

matthewdylan:
> All LGTM! It's kinda cool to see the Waco branding scaling this way. I support the approach of duplication for the sake of getting this out the door. Also if we're following the [rule of 3](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)), this is technically in the clear 😈
> 
> Two flags:
> 
> * not code-related: Could you ping Anna C. about this if you haven't already?
> * future change: Our thank you pages aren't really "product" aware, which has always been the case regardless of this PR. BUT I did notice that the waco one in particular refers to the "Waco Local News Initiative" which I think was just the name before the name. Fine to update that in a future PR IMO!

Can do and noted!

matthewdylan:
Thank you both for the quick reviews!!!

PR #1250, opened 2025-06-11
adding in donor wall query pieces from tt app

matthewdylan:
#### What's this PR do?
Moves over initial pieces from the donor wall in the TT app
* adding boto3 to requirements for library to push data to s3
* moving query functionality to Account class in npsp
* moving _push_to_s3 function to util
* adding celery task that can be triggered nightly to push to s3

We need to move the donor wall out of the TT app and this gets the process to a point where it should be able to export a json file of donor info to s3 for further perusal.

I'll deploy this to donations-staging and should be able to start a nightly batch job from there.

Doesn't need to be right now.

This currently grabs all donors for any TT newsroom (Texas or Waco). I'm currently working with our salesforce contractor to utilize newsroom-specific giving fields for this going forward.

https://airtable.com/appyo1zuQd8f4hBVx/tbloNZu8GkM52NKFR/viwS1XPty68eK4Ett/recNQphHKCt7YNFKJ?blocks=hide

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,3 +23,4 @@ slack-sdk==3.12.0
 pydantic==1.9.0
 pytest-mock==3.10.0
 pytest-recording==0.12.2
+boto3==1.19.7
--- a/server/app.py
+++ b/server/app.py
@@ -67,6 +67,7 @@ from .util import (
     send_slack_message,
     send_cancellation_notification,
     name_splitter,
+    _push_to_s3,
 )
 
 ZONE = timezone(TIMEZONE)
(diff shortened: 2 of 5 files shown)


fillerwriter:
The code itself looks fine to me. As I understand it, we'll use the `list_by_giving` method to pull a collection of donations to render on the page, correct?

If that's the case, let's move the destination branch to `feature/donor-wall` rather than `master` so that we're not adding anything to production before we're ready to ship. If there's another way we want to manage this instead (feature flags, iterative deploys), let me know.

PR #1255, opened 2025-06-30
setup scheduled export of donor list

matthewdylan:
#### What's this PR do?
Sets up the push_donor_list function to run on a schedule determined by an env var (BATCH_HOURS)

This creates an updated json file of donors for the last 365 days that we can ref instead of having to do a direct (and costly) salesforce call on the donor wall page.

I'll deploy this branch to staging when appropriate and test out the celery scheduling functionality (kind of hard to test that part locally).

Doesn't need to be.

None come to mind.

### DIFF
--- a/server/config.py
+++ b/server/config.py
@@ -38,7 +38,7 @@ CELERY_ALWAYS_EAGER = bool_env("CELERY_ALWAYS_EAGER")
 # deprecated:
 CELERYBEAT_SCHEDULE = {
     "every-day": {
-        "task": "server.batch.charge_cards",
+        "task": "server.app.push_donor_list",
         "schedule": crontab(minute="0", hour=BATCH_HOURS),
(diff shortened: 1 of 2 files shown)


server/config.py
```diff
@@ -38,7 +38,7 @@ def bool_env(val):
 # deprecated:
 CELERYBEAT_SCHEDULE = {
     "every-day": {
-        "task": "server.batch.charge_cards",
```
fillerwriter, line 41: `charge_cards` seems important, but I don't know if the system is handling recurring charges through some other system outside of itself (like Stripe recurring charges or something).

fillerwriter (approved):
Assuming `server.batch.charge_cards` shouldn't be handled somewhere else (or is outside of this PR), this looks good to me. 👍

PR #1265, opened 2026-06-05
Fix GTM and GA tags to be Austin Current's rather than Waco's

seanvanderaa:
Fixing GTM and GA tag to be Austin Current's rather than Waco's. New AC GA tag is routed through the GTM tag. The hard-coded GA tag is for the ad grant tracking.


matthewdylan (approved):
LGTM! Thanks!

